PR #1, opened 2014-11-06
now adds hosts by name instead of UUID

JamesBelchamber:
This adds the Host VMs by name instead of UUID, which is a lot cleaner.

### DIFF
--- a/virt/esx/esx.py
+++ b/virt/esx/esx.py
@@ -212,7 +212,7 @@ class Esx(virt.HypervisorVirt):
             return
 
         # Get list of host uuids, names and virtual machines
-        object_contents = self.RetrieveProperties('HostSystem', ['vm', 'hardware'], hostObjs)
+        object_contents = self.RetrieveProperties('HostSystem', ['vm', 'hardware', 'name'], hostObjs)
         for host in object_contents:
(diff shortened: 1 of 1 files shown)


virt/esx/esx.py
```diff
@@ -310,7 +312,7 @@ def getHostGuestMapping(self):
                     # Stopped machine doesn't have any uuid
                     if vm.uuid is not None:
                         l.append(vm.uuid)
-                mapping[host.uuid] = l
+                mapping[host.name] = l
```
alikins: is 'escxli system uuid get' the instanceUuid ( https://www.vmware.com/support/developer/converter-sdk/conv43_apireference/vim.vm.ConfigInfo.html ) ?

The uuid virt-who is getting is the VirtualMachineConfig.uuid instead.

nokernel: What if he simply replace the mapping, and send the hostname for the HOST mapped to VMs UUID?

rnovacek:
Commit f6bed0f introduces the ability to identify hypervisor by hostname. We're working on reporting more information about the hypervisor that will ease the usage even more, but that requires server-side changes (Satellite 6 and Subscription Asset Manager).

PR #3, opened 2015-01-05
Allow Running On Unregistered Systems

dgoodwin:
Instead of requiring a registered system identity to use as a credential for authenticating with Satellite/SAM, this PR adds support for optional rhsm_username and rhsm_password (or rhsm_encrypted_password) options in a /etc/virt-who.d/ config file. If found these will be used instead of the identity certificate.

Please let me know of anything I missed or need to do to whip this into shape.


dgoodwin:
Only implemented this for /etc/virt-who.d/ config files, as that seems like the best way forward. Straight CLI options and environment variables are not included. Let me know if this is a problem.

dgoodwin:
Re-added the ESX performance fix from PR #2  that got clobbered by sync with fedorahosted.

PR #6, opened 2015-01-23
I also needed a better way to Identify the ESX hosts

nokernel:
With my bad python I have added a configuration option VIRTWHO_ESX_REGNAME and parameter --esx-regname

This new option is defaulted to False.

When set to True or 0, it will create a different host/guest mapping for ESX.

Instead of :

{ uuid: [ virt, virt, virt, ...], ...

Will do
{ hostname: [ virt, virt, virt, ... ], ...


rnovacek:
Commit f6bed0f introduces the ability to identify hypervisor by hostname. We're working on reporting more information about the hypervisor that will ease the usage even more, but that requires server-side changes (Satellite 6 and Subscription Asset Manager).

PR #12, opened 2015-06-17
 Added configuration options for specifying per-host rhsm config

Ceiu:
- Hosts can now specify most, if not all, rhsm configuration options
  in their host config. The configuration follows the same format as
  the rhsm.conf, except each option is prefixed with "rhsm_".


config.py
```diff
@@ -121,16 +132,60 @@ def fromParser(self, name, parser):
         try:
             rhsm_password = parser.get(name, "rhsm_password")
         except NoOptionError:
+            rhsm_password = None
+
+        try:
+            rhsm_host = parser.get(name, "rhsm_host")
```
dgoodwin: So we're largely re-creating the [server] section of rhsm.conf here, with an rhsm_ prefix. As such we should probably keep property names exactly the same, so could we make this one rhsm_hostname.

config.py
```diff
@@ -121,16 +132,60 @@ def fromParser(self, name, parser):
...
+        except NoOptionError:
+            rhsm_host = None
+
+        try:
+            rhsm_ssl_port = parser.get(name, "rhsm_ssl_port")
```
dgoodwin: Just rhsm_port here perhaps for same reasoning?

config.py
Ceiu: Ahh, that one slipped my mind. Fixed.

tests/test_config.py
```diff
@@ -188,6 +244,16 @@ def testMultipleConfigFiles(self):
...
+rhsm_username=rhsm_admin1
+rhsm_password=rhsm_password1
+rhsm_hostname=host1
+rhsm_ssl_port=12341
+rhsm_prefix=prefix1
+rhsm_proxy_hostname=proxyhost1
+rhsm_proxy_port=43211
+rhsm_proxy_user=proxyuser1
+rhsm_proxy_password=proxypass1
+rhsm_insecure=1
```
dgoodwin: Just want to check, does this function ok if you don't specify all of these properties? I.e. do we fallback to the original rhsm.conf values if something isn't found?

dgoodwin:
Should probably document in the man pages, virt-who-config.5 I think.

dgoodwin:
Everything else looks good for what we needed of this thanks.

@rnovacek over to you for review now. :)

tests/test_config.py
Ceiu: Yes. As of the most recent commit, the values in rhsm.conf are used as the basis, and those provided in the host config override the rhsm.conf "defaults."

config.py
```diff
@@ -43,10 +43,13 @@ def parse_list(s):
     '''
     return reader([s]).next()
 
+# TODO: Not a huge fan of this. This whole class should be replaced by a dictionary or
+# dictionary-like class that provides a __missing__ implementation or "safe" getter.
```
rnovacek: Totally agree, it will rework it.

virt-who-config.5
```diff
@@ -37,6 +37,30 @@ Optional password to use to communicate with Subscription Asset Manager or Satel
 \fBrhsm_encrypted_password\fR
 Alternative to the rhsm_password option, encrypted password that is generated by virt-who-password(8) utility.
 .TP
+\frhsm_hostname\fR
+Optional hostname of the Subscription Asset Manager or Satellite server to use in place of the host defined in the system's rhsm.conf.
```
rnovacek: Please state explicitly it's for "Satellite 6 server".

manager/subscriptionmanager/subscriptionmanager.py
```diff
@@ -58,8 +58,11 @@ def readConfig(self):
         self.cert_file = os.path.join(consumerCertDir, cert)
         self.key_file = os.path.join(consumerCertDir, key)
 
-    def _connect(self, rhsm_username=None, rhsm_password=None):
+    def _connect(self, config = None):
         """ Connect to the subscription-manager. """
+
+        # Do we want to treat the rhsm_config as the default, and everything else as overrides, or
+        # should we only take from one set of configuration?
```
rnovacek: I would say that these option should override options from rhsm_config. So use rhsm_* option from virt-who config if defined and if not, use value from rhsm.conf.

rnovacek:
It looks really good, thanks. Please fix the "Satellite" -> "Satellite 6" in the man page and replace the question in the comment with statement and I'll merge it.

Ceiu:
Done. Also renamed the rhsm_ssl_port to rhsm_port for consistency with rhsm.conf

rnovacek:
Merged, thank you.

PR #13, opened 2015-06-24
Adds support for the new async report api

cnsnyder:
Added:
- Hypervisor class
     - this is simply a model (similar to Guest) of a hypervisor
     - this class includes fields to match the new api (including hostname)
     - Many of the changes in this branch are to do with the addition of this class

- Job class
    - This class is presently a simple model of the data contained in an async job

Changes:
-All virt backends now use the Hypervisor class in their getHostGuestAssociation methods
-SubscriptionManager.hypervisorCheckIn() now reformats the HostGuestAssociation depending on the connected candlepin's capabilities.
- The VirtWho class is now capable of executing tasks after a delay.
    - This functionality has been used to enable polling for the results without tying up the virtwho main
       loop completely

This PR requires a merge of the branch by the same name (csnyder/new_report_api) in python-rhsm to work properly .


virt/virt.py
```diff
@@ -77,6 +77,37 @@ def toDict(self):
...
+
+        'name': the hostname, if available
+        """
+        self.hypervisorId = hypervisorId
+        self.guestIds = guestIds
+        self.name = name
+
+    def toDict(self):
+        d = {
+            'hypervisorId': self.hypervisorId,
```
wottop: The json looks like this:
  "hypervisorId" : {
    "hypervisorId" : "uuid_0"
  }
It looks like this is missing a layer.

virt/virt.py
```diff
@@ -77,6 +77,37 @@ def toDict(self):
...
+            'guestIds': [g.toDict() for g in self.guestIds]
+        }
+        if self.name is not None:
+            d['name'] = self.name
+        return d
+
```
wottop: We are also looking to allow fact usage [sockets and what not] for the hosts. Not sure how/when they are to be collected on this end, but we should build the means to send them.

virt/virt.py
```diff
@@ -120,13 +151,14 @@ def __init__(self, config, assoc):
     @property
     def association(self):
         # Apply filter
+        # TODO fix the filtering to work with the new hypervisor class
```
wottop: TODO done?

manager/subscriptionmanager/subscriptionmanager.py
```diff
@@ -139,19 +140,61 @@ def sendVirtGuests(self, guests):
...
+                self.logger.info("Updated host %s with guests: [%s]",
+                                 updated['uuid'],
+                                 ", ".join(guests))
+            if (isinstance(result['created'], list)):
+                for created in result['created']:
+                    guests = [x['guestId'] for x in created['guestIds']]
+                    self.logger.info("Created host: %s with guests: [%s]",
+                                     created['uuid'],
+                                     ", ".join(guests))
+            result = resultData
```
wottop: You should report at least the count of unchanged.

virt/virt.py
cnsnyder: The hypervisor class is used for both but in the manager/subscriptionmanager.py Subscription Manager class the data to be sent is changed to the old format if necessary.

Re: the missing layer, I'll go change that :+1:

manager/subscriptionmanager/subscriptionmanager.py
```diff
@@ -139,19 +140,61 @@ def sendVirtGuests(self, guests):
...
         """ Send hosts to guests mapping to subscription manager. """
-
         serialized_mapping = {}
-        for host, guests in mapping.items():
-            serialized_mapping[host] = [guest.toDict() for guest in guests]
-        self.logger.info("Sending update in hosts-to-guests mapping: %s" % json.dumps(serialized_mapping, indent=4, sort_keys=True))
 
         self._connect(config)
+        self.logger.debug("Checking if server has capability 'hypervisor_async'")
+        is_async = self.connection.has_capability('hypervisors_async')
```
rnovacek: Stay backwards compatible with older versions of rhsm, for example:
    is_async = hasattr(self.connection, 'has_capability') and self.connection.has_capability('hypervisors_async')

virt/rhevm/rhevm.py
```diff
@@ -161,12 +165,9 @@ def getHostGuestMapping(self):
...
                 state = virt.Guest.STATE_UNKNOWN
 
-            mapping[host_id].append(virt.Guest(guest_id, self, state))
+            hosts[host_id].guestIds.append(virt.Guest(guest_id, self, state))
 
-        mapping_with_ids = {}
-        for host_id, vms in mapping.items():
-            mapping_with_ids[host_ids[host_id]] = vms
-        return mapping_with_ids
+        return {'hypervisors': [h for id, h in hosts.items]}
```
rnovacek: This raises an error (missing function call of the `.items`), it's this what you meant?
```python
return {'hypervisors': hosts.values()}
```

rnovacek:
I'm not quite sure that the background process is necessary. Given that the communication is async now, both calling hypervisorCheckIn and checking for the result should be pretty fast, right? And the main process is not doing anything in the meantime apart from moving items from one queue to another, if I understand the code correctly.

Am I missing something here?

virt/virt.py
cnsnyder: Yes, I've removed it and will update the PR along with other changes soon.

manager/subscriptionmanager/subscriptionmanager.py
cnsnyder: Thank you, adding that change

cnsnyder:
@rnovacek Indeed the main process is doing nothing but moving items from one queue to another in this case. In addition the manager process is presently doing very much the same thing. Many of the "jobs" that it is tasked with amount to placing a particular job on its out going queue at a particular time as most of the methods we actually want to be called are in virtwho itself. 

Given that, would it make more sense to move this waiting functionality to the virtwho class directly instead of spinning up a separate process?

cnsnyder:
At this point I believe all concerns have been addressed. Is there anything else that should be modified before this can be merged?

virtwho.py
```diff
@@ -25,20 +25,62 @@
...
 from daemon import daemon
-from virt import Virt, DomainListReport, HostGuestAssociationReport, ErrorReport
+from virt import Virt, DomainListReport, HostGuestAssociationReport, ErrorReport, Hypervisor
 from manager import Manager, ManagerError, ManagerFatalError
+from manager.managerprocess import ManagerProcess
+from manager.subscriptionmanager import SubscriptionManager
 from config import Config, ConfigManager, InvalidPasswordFormat
 from password import InvalidKeyFile
+from datetime import datetime, timedelta
+from heapq import *
```
rnovacek: Don't use * import, specify used imports only.

virtwho.py
```diff
@@ -83,7 +125,12 @@ def __init__(self, logger, options, config_dir=None):
         self.logger = logger
         self.options = options
         self.terminate_event = Event()
+
+        # Queue for getting events from virt backends
         self.queue = None
+        self.manager_in_queue = Queue()
```
rnovacek: Is this used? I don't see it.

virtwho.py
```diff
@@ -175,50 +240,67 @@ def run(self):
             return
 
         result = {}
+        report = None
         while not self.terminate_event.is_set():
             # Wait for incoming report from virt backend
             try:
-                report = self.queue.get(block=True, timeout=None)
+                report = self.queue.get_nowait()
```
rnovacek: This looks like busy waiting if I'm not mistaken. We'll just check for Jobs in the self.jobs in loop without any waiting. I think we should wait for new event in self.queue for some time and then check all the Jobs in the self.jobs.

rnovacek: I don't think it's necessary for each Job to have it's own timeout. Just create list of Jobs to be checked next time we're checking them.

virtwho.py
```diff
@@ -25,20 +25,62 @@
...
 from password import InvalidKeyFile
+from datetime import datetime, timedelta
+from heapq import *
 
 import log
 
 from optparse import OptionParser, OptionGroup, SUPPRESS_HELP
 
+# Default interval for jobs to be delayed
+DefaultJobInterval = 60
```
rnovacek: I think 60 seconds is quite a lot, esp. for first run with small amount of data. In previous version the submission was almost immediate, now it just 60 seconds even when no association is sent.

cnsnyder: Reports are still sent as soon as they are collected. This presently only affects the retrieval of the status of async jobs from candlepin. What would be more reasonable job timeout?

virtwho.py
cnsnyder: Having timeouts for each Job allows us to check and record the status of async jobs at some interval other than that of collecting the virt info from the virt backends. Is this unnecessary? If we remove the individual timeouts we would end up checking the results of an async call and deciding what to do as a result only when we collect new data. Which is the preferable behavior?

cnsnyder: I agree that for the case that it is currently used having timeouts for each job seems a little excessive. However, it will likely be very useful to have later when we implement rate-limiting.

cnsnyder:
@rnovacek I have removed timeouts from jobs entirely. Is there anything else that might need changing?

rnovacek:
It looks fine now, merging. Thanks for the contribution.

PR #14, opened 2015-07-02
Ensures jobs are removed after being executed

cnsnyder:


cnsnyder:
Good point. I can fix that.

PR #19, opened 2015-07-31
Updates virtwho logging to increase flexibility of the logging system

cnsnyder:
# Summary
The main change is that all loggers (returned by log.getLogger()) use QueueHandlers to send their log messages to an instance of QueueLogger. QueueLoggers run a thread that logs from a queue.
The handlers that would be added directly to a logger are instead added to the QueueLogger with a logging.Filter of the name of the original logger. As all log records are sent to one thread to be logged multiple loggers can write to one file or many.

A method of creating and or providing a single default instance of a QueueLogger is provided (see log.getDefaultQueueLogger()). This allows easy access to the same object from multiple places.

The default behavior is for a log file (in the default directory) to be created for each config provided to virtwho. These files are names are generated based on the name of the config.

## Options changes:
Using the new option '--log-per-config' causes all loggers to write to their own file.
You can change the name of the default single log file using the '--log-file' option followed by the name of the file.
You can change the directory all log files will be placed in by adding '--log-dir' followed by the absolute path of the directory.

## Environment variable changes:
Added three new environment variables, one for each of the new options:
- VIRTWHO_LOG_DIR
- VIRTWHO_LOG_FILE
- VIRTWHO_SINGLE_LOG_FILE

## Config Changes:
There are two new values that can be configured per virtualization backend:
- log_dir=/absolute/path/to/dir/where/this/virt/should/log/to
- log_file=name_of_file_to_log_to.log

The settings in the config will be used if available. If not the settings in the options will be used as default. If both an environment var and option are specified for the same value the environment var is used.

## Changes to the default logging behavior
- The default logging directory and file are the same '/var/log/rhsm/ and 'rhsm.log' as previously
- Without the '--log-per-config' option, virt-who defaults to logging to the default single file


rnovacek:
I agree that splitting the logs per config is useful, but we should stay backwards compatible. All the log messages should be copied also to rhsm.log as it used to. We state in every docs that the log is in /var/log/rhsm/rhsm.log file. I don't think we should change it. But we can add the per config logs anyway, but it would mean duplicated entries.

I think that by default (when user doesn't reconfig it), the log should go rhsm.log as it used to.

What do you think?

cnsnyder:
@rnovacek I was not really sure about whether to log per config by default or to log to the single file by default. It would probably make sense to maintain logging to a single file by default and change the options and environment variables to enable per config logging if so desired.

As for which file we should log to by default. I still think it would be useful to have a separate file and directory for virtwho logging by default. With regards to the docs, is there any reason (aside from tracking them down and making the changes) that we can't update them to reflect this change of log file and directory?

PR #20, opened 2015-08-06
Patch logging.FileHandler._open to prevent any actual file access

cnsnyder:
Some of the tests (involving FileHandlers) were trying to open various files on the file system.
This patch mocks out logging.FileHandler._open for those tests which should prevent any files from actually being opened.


rnovacek:
There is a fail with python2.6:

TypeError: super() argument 1 must be type, not classobj

https://travis-ci.org/virt-who/virt-who/jobs/74433229#L642

PR #21, opened 2015-08-19
Ensures VirtWho.terminate() and log.getDefaultQueueLogger().terminate…

cnsnyder:
…() are called when we need to exit


cnsnyder:
The checks that have failed have been fixed on another branch that also has a PR here: https://github.com/virt-who/virt-who/pull/20

cnsnyder:
Rebased on #20

PR #22, opened 2015-08-20
Fixed string interpolation issues in virt-who

vritant:
* fix issues where logging was not interpolating strings right.
* verified all strings with thwo %s are now with correct interpolation in the code base
* also verified there are no strings with more than two %s

### DIFF
--- a/manager/subscriptionmanager/subscriptionmanager.py
+++ b/manager/subscriptionmanager/subscriptionmanager.py
@@ -194,15 +194,15 @@ class SubscriptionManager(Manager):
                     self.logger.error("Error during update list of guests: %s", str(fail))
             for updated in resultData['updated']:
                 guests = [x['guestId'] for x in updated['guestIds']]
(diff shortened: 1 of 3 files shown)


vritant:
@rnovacek please take a look

rnovacek:
According to python logging module docs, "logger.info('message: %', something)" is correct. What is the benefit of having it change to normal string formatting instead of having logger handle that?

rnovacek:
The fix of formatting when throwing the exception if definitely correct, thanks.

vritant:
@rnovacek  you are correct that logging module supports args with string interpolation in that format, However, after conversation with @cnsnyder , we found that log.py delegation to logger breaks that feature.

rnovacek:
@vritant I think it better to fix it in the logger than having to remember not to use that feature. This commit fixed it for me: fd9e259d9dc0f02c5587daa99683a4b647b6bcc8

Is this solution acceptable?

vritant:
Closing PR after verifying the proper fix.

PR #24, opened 2015-09-11
Obey http response code 429

cnsnyder:
Virt-who now sends one report (that has changed) per the amount of seconds specified in the interval.

Upon receipt of a 429 reponse from the server (SAM or candlepin), virt-who will modify it's wait time based on the Retry-After header (if included, if not a default of 2) to the power of the number of 429s it has received in a row. This count is reset as soon as a report is sent without error.

### DIFF
--- a/tests/test_virtwho.py
+++ b/tests/test_virtwho.py
@@ -182,7 +182,8 @@ class TestOptions(TestBase):
         getLogger.return_value = sentinel.logger
         fromConfig.return_value.config.name = 'test'
         virtwho = VirtWho(self.logger, options, config_dir="/nonexistant")
-        config = Config("test", "esx", "localhost", "username", "password", "owner", "env")
+        config = Config("test", "esx", server="localhost", username="username",
+                        password="password", owner="owner", env="env")
         virtwho.configManager.addConfig(config)
         virtwho.queue = Queue()
         virtwho.queue.put(HostGuestAssociationReport(config, association))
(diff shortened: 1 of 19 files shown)


tests/test_virtwho.py
```diff
@@ -278,10 +283,32 @@ def fake_virts(logger, config):
         self.assertTrue(virtwho.reports[self.config.hash] == self.fake_report.hash)
         self.assertTrue(virtwho.reports[self.second_config.hash] == self.fake_domain_list.hash)
 
+#    @patch('log.getLogger')
```
rnovacek: Please, uncomment or remove this function.

virtwho.py
```diff
@@ -129,14 +130,20 @@ def __init__(self, logger, options, config_dir=None):
...
         # a heap to manage the jobs we have incoming
         self.jobs = []
-        # A dictionary of reports previously sent
+        # A dictionary of hashs of reports previously sent
         self.reports = {}
+        # a list of configs that have reports ready to send
+        self.configs_ready = []
+        # A dictionary of reports to send
+        self.reports_to_send = {}
+        self._429_count = 0
```
rnovacek: Please add a comment why is it call 429.

tests/test_virtwho.py
rnovacek: There is another commented function bellow.

rnovacek:
Otherwise it looks good, thanks.

PR #30, opened 2015-12-04
Xen plugin

wido-gg:
I have created a plugin for XenCenter. Where I work, we're RH customers and are using Satellite 6.1 with XenCenter. Since Xen is not supported, we needed this to work and since we managed to do so, we wanted to contribute it back. 

It work very similar to the esx plugin. I could write a Doc if you like. Regards


virt/xen/xen.py
```diff
@@ -0,0 +1,294 @@
+#!/usr/bin/env python
+
+import sys
+from time import time
+import XenAPI
+from XenAPI import Failure, NewMaster
+from collections import defaultdict, OrderedDict
+import virt
+
+class Guest(object):
```
rnovacek: Inherit this from virt.Guest and rename it to something like XenGuest.

rnovacek: Is it even necessary to have it here? What is the difference between it and Guest class from virt/virt.py module?

virt/xen/xen.py
```diff
@@ -0,0 +1,294 @@
...
+        ))
+
+        if self.hypervisorType is not None:
+            d['attributes']['hypervisorType'] = self.hypervisorType
+            if self.hypervisorVersion is not None or \
+               self.hypervisorVersion is not "":
+                d['attributes']['hypervisorVersion'] = self.hypervisorVersion
+        return d
+
+class Hypervisor(object):
```
rnovacek: Why it is here? It seems that the implementation is the same as Hypervisor class in virt/virt.py.

virt/xen/xen.py
```diff
@@ -0,0 +1,294 @@
...
+    def login(self):
+        # Login to server using given credentials
+        try:
+            # Don't log message containing password
+            self.session = XenAPI.Session(self.url)
+            self.session.xenapi.login_with_password(self.username, self.password)
+            self.logger.debug("I'm in XEN pools! with user %s" % self.username)
+
+        except NewMaster as nm:
+            self.logger.debug("Tengo un nuevo master?")
```
rnovacek: English please :)

virt/xen/xen.py
```diff
@@ -0,0 +1,294 @@
...
+            try:
+                self.session = XenAPI.Session('http://%s' % nm.new_master())
+                self.session.xenapi.login_with_password(self.username, self.password)
+                self.logger.debug("I'm in XEN pools with new master %s! with user %s" % (nm.new_master(), self.username))
+
+            except:
```
rnovacek: Do not use plain 'except', you should specify what exceptions you expect here or at lease use 'except Exception:'.

virt/xen/xen.py
```diff
@@ -0,0 +1,294 @@
...
+                self.logger.exception("Unable to login to XENserver %s" % self.url)
+                raise
+
+        except:
```
rnovacek: Same as above, do not use plain except.

virt/xen/xen.py
```diff
@@ -0,0 +1,294 @@
...
+
+                guests.append(virt.Guest(
+                        uuid= vm["uuid"],
+                        virt= self,
+                        state= state,
+                        hypervisorType=None
+                    )
+                )
+
+            mapping[record["hostname"]]=guests
```
rnovacek: Please use new format of getHostGuestMapping (sorry, its not defined anywhere), it's a dict with following format:
``python
mapping = {
    'hypervisors': [
        Hypervisor(hypervisorId=uuid, guestIds=guests, name=name, facts=facts)),
        ...
    ]
}
```

Comment (rnovacek) [inline virt/xen/xen.py]: Would it be possible to support hypervisor_id feature and identify host based on either UUID, hostname or hardware UUID? Does XEN provide this information via its API?

File: virt/xen/xen.py [hunk 7]
```diff
@@ -0,0 +1,294 @@
...
+        return mapping
+
+    def _run(self):
+        self._prepare()
+
+        self.hosts = defaultdict(Host)
```
rnovacek: self.hosts and self.vms are not used, right?

virt/xen/xen.py
```diff
@@ -0,0 +1,294 @@
...
+            end_time = time()
+
+            if self._oneshot:
+                break
+
+            self.logger.debug("Waiting for XEN changes")
+
+        #self._cancel_wait()
+
+class Host(dict):
```
rnovacek: Are these classes used anywhere?

virt/xen/xen.py
wido-gg: I have worked based on version 14.1. I *think* (out of my memory) these kind of things changed in version 15. I'll take a look, but I may need some mentoring

virt/xen/xen.py
wido-gg: Xen provides UUID and hostname. I'm not sure what's the difference between uuid and hardware uuid and I'm not seeing that in the xen documentation

virt/xen/xen.py
wido-gg: I really really really tried to do a pull request against tag:14, but ended up doing it against master. virt.Guest is not present in version 14, which is what I'm using (is what RH has as stable in RH7). There are things I half-backported, most I thought would be useful, and some I wasn't entirely sure what they do. I do watn, however, to clean the code and make it as good as possible for this project.

rnovacek: Sure, it's definitely a good start. If you could rebase it against current master and solve the issues I found, I think it will be almost ready to merge.

virt/xen/xen.py
rnovacek: Support for UUID and hostname is enough. We'll just note in the docs that hwuuid is not supported.

wido-gg: What does 'facts' mean? This would be the new code (but properly indented):

mapping['hypervisors']= Hypervisor (hypervisorId=hypervisor_record["uuid"], guestIds=guests, name=hypervisor_record["hostname"], facts="")

rnovacek: 'Facts' is a dictionary with additional information about the hypervisor, only reported data so far is number of sockets of the hypervisor. Feel free to set `facts=None` for now.

rnovacek:
This pull request is obsoleted by PR#31.

PR #38, opened 2016-12-16
Add config directives to allow filtering hosts according to regex not just string matches

chr15p:
I have a customer with a single vcentre infrastructure manageing a large number of esx servers. They restrict the guests by esx to linux on hosts names esxlin*, windows on esxwin* and vdi on esxvdi.  Because of this their exclude_hosts list is now approximately 40 hosts long, which is unmanageable. 

This patch introduces a new config directive exclude_hosts_regex that acts exactly like exclude_hosts but instead takes comma separated list of (python) regular expression to allow users to filter automatically in a neater and more manageable fashion. i.e. exclude_hosts_regex="esxwin.*","esxvdi.*"

### DIFF
--- a/virt-who-config.5
+++ b/virt-who-config.5
@@ -122,6 +122,12 @@ Only hosts which uuid (or hostname or hwuuid, based on \fBhypervisor_id\fR) is s
 \fBexclude_hosts\fR
 Hosts which uuid (or hostname or hwuuid, based on \fBhypervisor_id\fR) is specified in comma-separated list in this option will \fBNOT\fR be reported. Put the value into the double-quotes if it contains special characters (like comma). \fBexclude_host_uuids\fR is deprecated alias for this option.
 .TP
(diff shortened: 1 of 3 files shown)


cnsnyder:
@chr15p Is there a rhbz associated with this?

Looks good to me. :+1:  Thank you for the contribution!

chr15p:
Hi,

No, I didn't raise a bz, I guess I can if you want but it was more a "I
wonder how easy it would be to..." thing then a bug :-)

Glad you like it!

chris

cnsnyder:
@chr15p Hey Chris, Thanks again for your contribution. I think that bz matches this perfectly. One final request, can you please add the bz to the beginning of the commit message? Thank you!

I'll merge this in once that is done.

chr15p:
Done.  (Hopefully, assuming I've got the right mystical incantations to bend git to my will!)

PR #39, opened 2017-01-04
1369107: Only load files with .conf extension

pcreech:


cnsnyder:
:+1:  LGTM. Thank you for the contribution.

PR #40, opened 2017-01-09
1392390 Fix default interval handling

pcreech:


virtwho/parser.py
```diff
@@ -154,12 +154,14 @@ def parseOptions():
...
-    try:
-        if env and int(env) >= MinimumSendInterval:
-            options.interval = int(env)
-    except ValueError:
-        logger.warning("Interval is not number, ignoring")
+        try:
+            env = int(env.strip().lower())
+            if env >= MinimumSendInterval:
+                options.interval = env
+            elif env <= MinimumSendInterval:
```
cnsnyder: No need for '=' in both cases is there?

pcreech: Correct.  Just a slip on inverting the logic

PR #41, opened 2017-01-13
1400431: Fix AttributeError where val is missing

pcreech:


tests/test_esx.py
```diff
@@ -207,3 +207,145 @@ def test_proxy(self):
         self.assertRaises(VirtError, self.run_once)
         self.assertIsNotNone(proxy.last_path, "Proxy was not called")
         self.assertEqual(proxy.last_path, 'localhost:443')
+
+    def test_applyHostSystemUpdate_AttributeError(self):
+        changeSet = Mock(spec=['op', 'name'])
```
cnsnyder: Non_nacking suggestion: Might it be clearer what object returned from the server if this var were renamed from "changeSet"  to "change".

PR #43, opened 2017-01-17
Adds --password option to virt-who-pasword

tstrachota:
virt-who-password currently reads input only interactively via getpass
which prevents it from being used in scripts.

This patch adds fixes the issue by adding options -p and --password.


tstrachota:
@cnsnyder ping, may I ask you for review?

barnabycourt:
@pcreech would you mind taking a look?

pcreech:
Looks good @tstrachota 

Merging

PR #44, opened 2017-01-18
1405967: Filter host via glob or regex

pcreech:


cnsnyder:
@pcreech One last thing I meant to say before approval: Could you remove the 'BZ' from the beginning of the commit message and follow the bug number by a ':'?

e.x. '1405967: Filter host via glob or regex'

cnsnyder:
@pcreech I'm looking for a change to the commit message. Looks like changing the title on github doesn't affect the commit message. The reason I ask for this change is to ensure that when building virt-who using tito, tito can check the state of the bugs referenced in the commits.

We can always look into modifying tito to check for commits matching another format.

pcreech:
Ah, yeah.  Will fix soon

pcreech:
Updated commit message

PR #45, opened 2017-02-02
Move from using processes to threads

cnsnyder:
This is step 1 of the interval design.
As virt-who does not need to take advantage of multiple physical cores,
we have decided to move from using processes to threads.
We consider it more maintainable and reliable to get threads to stop
than processes.

Update tests to use threading where necessary

Fix patching to refer to the right Queue class

Update assertions to ensure the proper termination method is called

### DIFF
--- a/virtwho/virt/virt.py
+++ b/virtwho/virt/virt.py
@@ -23,7 +23,7 @@ import time
 import logging
 from operator import itemgetter
 from datetime import datetime
-from multiprocessing import Process, Event
+from threading import Thread, Event
 import json
 import hashlib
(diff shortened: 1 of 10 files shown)


virtwho/virt/virt.py
```diff
@@ -378,8 +378,6 @@ def run(self):
         '''
         self.logger.debug("Virt backend '%s' started", self.config.name)
         # Reset the signal handlers, we'll handle them only in the main thread
```
pcreech: After looking over the code, looks like this comment can be removed as well

PR #47, opened 2017-02-06
Step 2 of the interval design: Implement a threadsafe datastore

cnsnyder:
This is the implementation of step 2 of the interval design.

This step depends on all prior steps.

This PR adds a threadsafe datastore into the virt-who codebase. The datastore (as of this PR) is not in use in virt-who, it is to be used in subsequent steps of the design. More specifically, this datastore will be used to allow "destination" threads to consume reports produced by one or more "source" threads.

Related trello card: https://trello.com/c/HvuaQygd/275-interval-design-step-2-create-datastore
Design doc for the overall design: https://docs.google.com/document/d/1AV-eEp1HeemTHIiigGsUBDFK3W1XWMC1flSt9jrxek4/edit

### DIFF
--- a/tests/test_datastore.py
+++ b/tests/test_datastore.py
@@ -1,0 +1,140 @@
+from base import TestBase
+from mock import sentinel, patch, MagicMock
+from threading import Lock
+from virtwho.datastore import Datastore
+
+
+class TestDatastore(TestBase):
+
+    def setUp(self):
+        pickle_patcher = patch('virtwho.datastore.pickle')
+        self.mock_pickle = pickle_patcher.start()
+        self.mock_pickle.dumps.return_value = sentinel.pickled_value
+        self.mock_pickle.loads.return_value = sentinel.unpickled_value
+        self.addCleanup(pickle_patcher.stop)
+
+        lock_patcher = patch('virtwho.datastore.Lock')
+        mock_lock_class = lock_patcher.start()
+        mock_lock_instance = MagicMock(spec=Lock())
+        mock_lock_class.return_value = mock_lock_instance
+        self.mock_lock = mock_lock_instance
+        self.addCleanup(lock_patcher.stop)
+
+    def mock_test_data(self, datastore, **kwargs):
+        # Sets the datastore to contain the keys and values of the kwargs
+        # in a way that does not use the put method of the datastore
+        # Returns the datastore instance as modified, the test internal
+        # datastore
+        mock_internal_datastore = MagicMock(wraps=dict(kwargs))
+        datastore._datastore = mock_internal_datastore
+        return datastore, mock_internal_datastore
+
+    def test_get_uses_lock(self):
+        # Ensure a lock is acquired before returning a particular entry in the
+        # store
+
+        # Ensure that a lock is acquired before updating a particular entry
+        # in the store.
+        # These assertions assume the lock is used as a context manager
+        # and that the lock (when used as a context manager) is acquired by
+        # the calling thread on __enter__ and released by the calling thread
+        # on __exit__
+        datastore, mock_internal_datastore = self.mock_test_data(Datastore(),
+                                                                 test_item=sentinel.test_value)
+
+        def assert_internal_store_unchanged(*args, **kwargs):
+            # Assert there have been no accesses of the internal datastore
+            mock_internal_datastore.__getitem__.assert_not_called()
+            mock_internal_datastore.__setitem__.assert_not_called()
+
+        def assert_internal_store_accessed(*args, **kwargs):
+            mock_internal_datastore.__getitem__.assert_called_once_with(
+                    "test_item")
+            mock_internal_datastore.__setitem__.assert_not_called()
+
+        self.mock_lock.__enter__.side_effect = assert_internal_store_unchanged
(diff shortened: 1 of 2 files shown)


tests/test_datastore.py
```diff
@@ -0,0 +1,224 @@
...
+
+class TestDatastore(TestBase):
+
+    def mock_copy(self, item):
+        # Return the item with some predictable modification
+        # Used to prove that the result of copy method is used in the right
+        # places
+        return item, item
+
+    def assert_equivalent(self, item1, item2):
```
cnsnyder: Self comment, this might be absurd. I could be convinced to trust that any object we'd like to test equivalence of has a sane __eq__ implementation method.

pcreech: It can be considered rather complex, where we could test for standard equivalence.  Are we anticipating a need for storing complex custom objects?  Or standard python types and primitives?  if it's the latter, we can probably assume a sane "\_\_eq__" implementation

cnsnyder: Khowell had a good idea that I think I'll go ahead and implement for this PR. That being to rely on the objects being put into the store having a 'copy' or 'clone' method that will return a copy of the object for storage.

Alternately, we could have the objects to be put in the store implement an interface that requires a serialize and deserialize method. We could serialize the objects on the way in to the store and deserialize them on the way out.

In either of the above case we can test that the implemented method on the objects being copied, serialized or deserialized produce suitable copies. Then all we need to see is that the datastore uses the right method on the way in and out.

cnsnyder:
@pcreech I've pushed a few updates to this PR. The new impl only allows "Cloneable" objects to be placed into the store and only returns a clone of the objects placed into the store.

The implementation of Cloneable.clone by default is to perform a deep copy of self.
I've added a few tests to show that the cloned objects behave the way we'd like.
I've removed a few tests (and reworked some) for the datastore.

The commits are still separate to make removing and modifying any or all of them easy.

Let me know if you like this version better than the original and I'll update the PR accordingly (squash if liked, remove / rework if not).

virtwho/datastore.py
```diff
@@ -0,0 +1,68 @@
...
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""
+
+import cPickle
+from threading import Lock
+
+
+
```
pcreech: flake8: Too many lines

virtwho/datastore.py
```diff
@@ -0,0 +1,68 @@
...
+        underlying datastore not have an item for the given key.
+
+        @raises KeyError: A KeyError is raised when the underlying datastore
+        has no item for the given key and a default has not been provided.
+        """
+        with self._datastore_lock:
+            try:
+                item = cPickle.loads(self._datastore[key])
+                return item
+            except KeyError as e:
```
pcreech: flake8: Unused variable 'e'

SHould be able to omit 'as e' here

virtwho/datastore.py
```diff
@@ -0,0 +1,70 @@
...
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""
+try:
+    import cPickle as pickle
+except ImportError:
+    import pickle
+from threading import Lock
+
+
+
```
pcreech: Flake8: Too many blank lines

virtwho/datastore.py
```diff
@@ -0,0 +1,70 @@
...
+        underlying datastore not have an item for the given key.
+
+        @raises KeyError: A KeyError is raised when the underlying datastore
+        has no item for the given key and a default has not been provided.
+        """
+        with self._datastore_lock:
+            try:
+                item = pickle.loads(self._datastore[key])
+                return item
+            except KeyError as e:
```
pcreech: Flake8: local variable 'e' is assigned but never used

except KeyError: should work here (tested while reviewing)

tests/test_datastore.py
```diff
@@ -0,0 +1,141 @@
...
+        self.addCleanup(pickle_patcher.stop)
+
+        lock_patcher = patch('virtwho.datastore.Lock')
+        mock_lock_class = lock_patcher.start()
+        mock_lock_instance = MagicMock(spec=Lock())
+        mock_lock_class.return_value = mock_lock_instance
+        self.mock_lock = mock_lock_instance
+        self.addCleanup(lock_patcher.stop)
+
+
```
pcreech: flake8: too many blank lines

tests/test_datastore.py
```diff
@@ -0,0 +1,141 @@
...
+        datastore.put("test_item", "test_item")
+        self.mock_lock.__enter__.assert_called_once()
+        self.mock_lock.__exit__.assert_called_once()
+
+    def test_put_uses_pickle_dumps(self):
+        # Ensure that put uses the return value of pickle.dumps
+        test_item = "test_value"
+        test_key = "test_item"
+        datastore, mock_internal_ds = self.mock_test_data(Datastore(),
+                                                test_item=test_item)
```
pcreech: flake8: under-indented for visual indent

pcreech:
LGTM: Thanks!

PR #48, opened 2017-02-07
1415497: Support rhev4 auto detection and usage

pcreech:


virtwho/virt/rhevm/rhevm.py
```diff
@@ -71,18 +71,73 @@ def __init__(self, logger, config):
...
-        self.clusters_url = urlparse.urljoin(self.url, "api/clusters")
-        self.hosts_url = urlparse.urljoin(self.url, "api/hosts")
-        self.vms_url = urlparse.urljoin(self.url, "api/vms")
-
         self.auth = HTTPBasicAuth(self.config.username, self.config.password)
 
+    @property
+    def clusters_url(self):
+        if not hasattr(self, '_clusters_url'):
+            if self.major_version == '4' and 'ovirt-engine' not in self.url:
```
cnsnyder: Could we do the version check on init and update self.url to include the necessary prefix (for the old rhev set prefix to 'api/', for new set prefix to 'ovirt-engine/api/') ?

tests/complex/data/rhevm/rhev3_api.xml
```diff
@@ -0,0 +1,79 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<api>
+    <link href="/ovirt-engine/api/capabilities" rel="capabilities"/>
```
pcreech: This is xml taken directly from a rhev 3.6 installation.  It will listen on the older url of $hostname/api, but the 'official' url is $hostname/ovirt-engine/api

virtwho/virt/rhevm/rhevm.py
```diff
@@ -71,18 +71,73 @@ def __init__(self, logger, config):
...
+
     def get(self, url):
         """
         Call RHEV-M server and retrieve what's on given url.
         """
         try:
-            response = requests.get(url, auth=self.auth, verify=False)
+            if self.major_version == '4':
+                headers = dict()
+                headers['Version'] = '3'
```
pcreech: This is taken from the Version 3 api docs of Rhev4.  They approached versioning the api in different ways, and the most transparent way I could think of to utilize this is to add the 'Version: 3' header as suggested by the [docs](https://access.redhat.com/documentation/en/red-hat-virtualization/4.0/paged/version-3-rest-api-guide/chapter-3-rest-api-quick-start-example)

virtwho/virt/rhevm/rhevm.py
pcreech: I was having issues with mocking items by doing the version check on init.  Since the version check was reaching out and doing a request.get, it was creating complicated testing patterns.  I switched to using properties based on khowell's suggestion.

With this modification, I can just set 'self._major_version' on the object in the unit test setup and ensure that we short-circuit the version testing logic for those test.

virtwho/virt/rhevm/rhevm.py
cnsnyder: Is there any where we can get this version number when we determine the major version number in (getVersions)?

virtwho/virt/rhevm/rhevm.py
cnsnyder: I'd like to request that we move the logic of determining the full url for access to a particular collection into it's own method. In this way should there ever be a change as to how we determine what the url should be, we need only change it in one place.

It might also make sense to set the value of each of these attributes ({clusters,hosts,vms}_url) in the prepare method.

virtwho/virt/rhevm/rhevm.py
```diff
@@ -71,18 +71,72 @@ def __init__(self, logger, config):
...
+        if self.major_version == "4":
+            api_base = 'ovirt-engine/api/'
+        else:
+            api_base = 'api/'
+
+        self.clusters_url = urlparse.urljoin(self.url, api_base + clusters_endpoint)
+        self.hosts_url = urlparse.urljoin(self.url, api_base + hosts_endpoint)
+        self.vms_url = urlparse.urljoin(self.url, api_base + vms_endpoint)
+
+    def getVersion(self):
```
cnsnyder: Can we change the new method names to pep8 standard of lowercase separated by '_' where necessary?
(e.g. "get_version")

We can make a separate task to go back through the code base and perform this change on the rest of the functions and methods.

virtwho/virt/rhevm/rhevm.py
```diff
@@ -71,18 +71,72 @@ def __init__(self, logger, config):
...
+        self.prepared = False
+
+    def prepare(self):
+        if not self.prepared:
+            if not hasattr(self, 'major_version'):
+                self.getVersion()
+            self.buildUrls()
+            self.prepared = True
+
+    def buildUrls(self):
```
cnsnyder: Same request as for getVersion (see below).

virtwho/virt/rhevm/rhevm.py
```diff
@@ -71,18 +71,72 @@ def __init__(self, logger, config):
...
         """
         Call RHEV-M server and retrieve what's on given url.
         """
         try:
-            response = requests.get(url, auth=self.auth, verify=False)
+            if self.major_version == '4':
+                # If we are talking to a Rhev4 system, we need to specifically request
+                # the Rhev 3 version of the api.  To minimize code impact, we do this
+                # by setting a 'Version' header, as outlined in Rhev 4's "Version 3
+                # REST API GUide"
```
cnsnyder: Should "GUide" be all caps?

pcreech: Based on the spelling of the title of the document in reference, no it should not.... but the U should be lowercase.  nice catch

cnsnyder:
LGTM :+1: 

Thank you!

PR #49, opened 2017-02-15
Update ConfigManager to produce destination and source mappings.

cnsnyder:
This is step 3 of the interval design. The configuration is now
parsed into unique destinations and sources. Historical Config
objects are treated as sources and the destination information
is pulled out of them and tracked. In this way the main thread
will know how many destination/source threads (and of what type) to
start.

### DIFF
--- a/virtwho/config.py
+++ b/virtwho/config.py
@@ -87,6 +87,112 @@ class NotSetSentinel(object):
     pass
 
 
+class Info(object):
+    """
+    A class containing keys with values.
+    If the kwargs passed to the init of this class do not include all
+    required_args, init will fail with a ValueError.
+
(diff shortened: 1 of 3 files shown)


virtwho/config.py
```diff
@@ -87,6 +87,118 @@ class NotSetSentinel(object):
...
+            if isinstance(item, list):
+                to_hash.append(tuple(item))
+            else:
+                to_hash.append(item)
+        return hash(tuple(to_hash))
+
+    def _set_option(self, key, value):
+        self._options[key] = value
+
+    def _get_option(self, key):
```
pcreech: Is this getter and the above setter needed, or can they be replaced with their implementation?

pcreech:
LGTM, thanks!

PR #51, opened 2017-02-27
Implement the Destination Thread and friends

cnsnyder:
This PR changes a number of things. Firstly common functionality between the Virt class (and subclasses) and what would be the Destination Thread class has been pulled out and put into a new super class for both IntervalThread. Secondly, I have implemented a DestinationThread class that should be able to use any existing manager to send reports gathered from the datastore (by name) to the right endpoints. Thirdly, tests have been added for the new DestinationThread (and updated for the Virt subclasses).

With any luck, this will be the second to last PR before the interval design will be functional.

### DIFF
--- a/tests/test_esx.py
+++ b/tests/test_esx.py
@@ -35,12 +35,14 @@ class TestEsx(TestBase):
     def setUp(self):
         config = Config('test', 'esx', server='localhost', username='username',
                         password='password', owner='owner', env='env')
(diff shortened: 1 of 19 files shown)


cnsnyder:
I will squash these commits down to one once this has been reviewed, modified if necessary, and accepted.

pcreech:
LGTM, Thanks for the contribution!

PR #52, opened 2017-03-06
1365255: Fix virt-who memory leak

pcreech:


cnsnyder:
We'll wait to see if this is the cause of the issue before delivering on it. (Will use the additional logging to help determine if this is the cause.)

PR #54, opened 2017-03-14
Add logging to esx add command

pcreech:
We can use this to intercept add command key-value pairs to understand
whats going on.


cnsnyder:
Will merge post-branch, closing until then

PR #55, opened 2017-03-15
Update the spec file for builds on more downstream platforms

cnsnyder:
Platforms that do not use systemd we will assume also will not
understand the macro %{__python2} or %{__python2_sitelib}


kahowell:
LGTM!

PR #57, opened 2017-03-24
Bring it all together with interval threads!

cnsnyder:
This PR modifies and utilizes the other building blocks from the other interval_* PRs to actually start using destination threads as per the design.

Note that this branch represents the minimum viable implementation (from my perspective) of the interval design and that much of what is contained herein will be changed in future work on virt-who configuration.

The ways to test this are numerous. I would suggest running this branch from source with different test configurations and options as you normally would with virt-who prior to the interval design and looking for any difference in functionality. The primary expected difference is that virt-who will now send all reports on the interval for each owner/env (for satellite 6 / hosted) and each configured satellite 5 instance per interval. In the case of hypervisor checkins for satellite 6 reports are now batched into one API call (one larger check in).

Post review: I will squash the many commits into something more reasonable and add a more useful summary of changes as the commit message for the single commit.

### DIFF
--- a/tests/complex/virtwhotest.py
+++ b/tests/complex/virtwhotest.py
@@ -1,4 +1,6 @@
 
+import os
+import signal
 import sys
 import socket
 from Queue import Empty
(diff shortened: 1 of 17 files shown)


pcreech:
LGTM!

PR #59, opened 2017-03-31
1436811: Send reports on the interval per destination

cnsnyder:
This set of changes allows virt-who to send all reports (batched where possible) on the interval for each destination. Reports for different destinations no longer block each other and cause check in times to increase. This allows virt-who to be set up more effectively for large organizations having some set of hypervisors that need to be reported to more than one organization or environment or even more than one satellite!


pcreech:
LGTM!

PR #62, opened 2017-04-12
Do not join threads not started, fix up fake backend

cnsnyder:
These are some bits that needed a bit of love post interval design.
The fake virt backend already had a method _get_data that was overriding the new super classes version of the method.
We certain threads fail to run properly in oneshot mode, it was possible to fail with a runtime exeception (and stall the process). This was due to an attempt to join a thread not started. This PR fixes that. In future builds, please add this as part of the RFE for the interval bits.

### DIFF
--- a/virtwho/executor.py
+++ b/virtwho/executor.py
@@ -139,7 +139,8 @@ class Executor(object):
     def terminate_threads(threads):
         for thread in threads:
             thread.stop()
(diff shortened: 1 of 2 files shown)


kahowell:
Ack, with the understanding that we have plans to clean up the threading more in the future. :smile:

PR #65, opened 2017-04-18
1442337: Send updates immediately the first run

cnsnyder:
To restore some historical behaviour that was lost during the implementation of the interval design, this PR causes virt-who to send all known reports as soon as it has received something from each source on the initial run.

With out this PR you must wait one interval before virt-who sends anything.

### DIFF
--- a/virtwho/virt/virt.py
+++ b/virtwho/virt/virt.py
@@ -394,9 +394,10 @@ class IntervalThread(Thread):
                     self._internal_terminate_event.set()
                     return
 
+                if has_error:
+                    self._send_data(ErrorReport(self.config))
+
                 if self._oneshot:
-                    if has_error:
-                        self._send_data(ErrorReport(self.config))
                     self.logger.debug("Thread '%s' stopped after running once",
                                       self.config.name)
                     self._internal_terminate_event.set()
@@ -449,6 +450,7 @@ class DestinationThread(IntervalThread):
         """
         if not isinstance(source_keys, list):
             raise ValueError("Source keys must be a list")
+        self.is_initial_run = True
         self.source_keys = source_keys
         self.last_report_for_source = {}  # Source_key to hash of last report
         self.options = options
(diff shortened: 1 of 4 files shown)


cnsnyder:
[1442337 - Failed to run virt-who for all the hypervisors with error message: "No report available for source"](https://trello.com/c/IbR8kMAw/361-1442337-failed-to-run-virt-who-for-all-the-hypervisors-with-error-message-no-report-available-for-source)

virtwho/virt/virt.py
```diff
@@ -492,6 +497,34 @@ def _get_data(self):
             reports[source_key] = report
         return reports
 
+    def _get_data_initial(self):
```
kahowell: This logic feels very similar to that of `DestinationThread._get_data`. What do you feel about abstracting the common bits out to a third method? Something like

```python
    def _get_data_common(self, source_keys):
        reports = {}
        for source_key in self.source_keys:
            report = self.source.get(source_key, NotSetSentinel)

            if report is None or report is NotSetSentinel:
                self.logger.debug("No report available for source: %s" %
                                  source_key)
                continue
            if report.hash == self.last_report_for_source.get(source_key, None):
                self.logger.debug('Duplicate report found, ignoring')
                continue
            reports[source_key] = report
        return reports
```

Then `_get_data` after the inital run logic branch can be simply `_get_data_common(self.source_keys)` and then `_get_data_initial` can contain only the waiting logic and the call to the common code... Thoughts?

virtwho/virt/virt.py
```diff
@@ -492,6 +497,34 @@ def _get_data(self):
...
+        @return: dict
+        """
+        reports = {}
+        while not reports and not self.is_terminated():
+            source_keys_remaining = list(self.source_keys)
+            time_waited = 0
+            while len(source_keys_remaining) > 0 and time_waited < self.interval and not self.is_terminated():
+                for source_key in source_keys_remaining:
+                    try:
+                        report = self.source.get(source_key)
```
kahowell: I noticed that unlike non-initial `_get_data`, here you do not check for `report is None or report is NotSetSentinel`... why is that?

virtwho/virt/virt.py
```diff
@@ -492,6 +497,34 @@ def _get_data(self):
...
+                        reports[source_key] = report
+                        source_keys_remaining.remove(source_key)
+                    except KeyError:
```
kahowell: Are you sure this is necessary? Looking at it, I thought initially it was for duplicate `source_key`(s), but remove will only remove one instance... If it is necessary, please leave a comment as to why.

virtwho/virt/virt.py
```diff
@@ -492,6 +497,34 @@ def _get_data(self):
...
+                        continue
+                if len(source_keys_remaining) > 0:
+                    sleep_time = 1
```
kahowell: Any significance to 1 second? If you have a specific reason, please document. No big deal if not, since 1 second seems reasonable.

virtwho/virt/virt.py
cnsnyder: I generally like the idea although I would say I think the log messages should be a little different for the initial checks vs the regular checks. I would also say that I don't want to be checking for duplicates etc in the initial run. I could remove that logic from this loop, but then that would mean going through all the items in the list returned from the common method in the non-initial checks to remove any that a duplicate.

virtwho/virt/virt.py
cnsnyder: I do not do this here because I had no interest in knowing whether the result I get back is nothing. The intention was to consider the item returned (so long as it exists) for whatever it is and consider us unblocked by that source (as we now have a value to try to send). Granted None will not result in anything being sent. Is it worth discarding None and spinning waiting for a Non-none value for all sources?

cnsnyder: I should also mention, the intention is for any source that has erred to place a report indicating such failure into the datastore. Even in the case of failure None should not be placed into the datastore. I suppose that does not mean we should not check for it.

Alright you've convinced me.

virtwho/virt/virt.py
cnsnyder: So as to not check repeatedly for the same key from the datastore. Why check for sources we have already gotten information for?

virtwho/virt/virt.py
cnsnyder: There is no specific reason.

kahowell: :+1:

cnsnyder:
@kahowell I have updated the PR to address some of your comments above. Hopefully this latest attempt more matches what is expected.

PR #66, opened 2017-05-09
1369107: Update docs and log message to show the *.conf requirement

cnsnyder:
These updates are related to the implementation of the RFE bug 1369107.
Herein are changes to the documentation as well as one of the log messages to help communicate the new requirement for config files to end in '.conf'.

### DIFF
--- a/virtwho/main.py
+++ b/virtwho/main.py
@@ -161,7 +161,8 @@ def main():
             exit(1, err)
         # In order to keep compatibility with older releases of virt-who,
         # fallback to using libvirt as default virt backend
-        logger.info("No configurations found, using libvirt as backend")
+        logger.info("No configurations found (are there any '.conf' files in /etc/virt-who.d?), "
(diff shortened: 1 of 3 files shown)


virtwho/main.py
```diff
@@ -161,7 +161,8 @@ def main():
             exit(1, err)
         # In order to keep compatibility with older releases of virt-who,
         # fallback to using libvirt as default virt backend
-        logger.info("No configurations found, using libvirt as backend")
+        logger.info("No configurations found (are there any '.conf' files in /etc/virt-who.d?), "
```
barnabycourt: Would it be useful to differentiate directly between no files in virt-who.d/ and having files that don't end in .conf?

cnsnyder: Yes I think so. I'll update the PR accordingly.

cnsnyder: PR updated accordingly!

PR #67, opened 2017-05-10
1448267: Fix polling behavior for oneshot, CTRL-C, 429 responses

cnsnyder:
This PR fixes the behaviour of polling when in oneshot mode, when CTRL-C is pressed, and when a http 429 response is received.

In oneshot mode, virt-who will make one attempt to report on each source at most. If any exception occurs during communication (including http 429), virt-who will move on. This is restoring the historical behaviour of virt-who oneshot mode.

When CTRL-C is pressed (to kill the virt-who process) virt-who now exits without polling or making additional connections. Prior to this PR, virt-who would start polling as fast as it could until it was successful.

When an http 429 is received virt-who will check for the existence of a Retry-After header. If provided and more than the minimumSendInterval (currently 60s) then that value will be used as the wait time until the request is retried. If not provided (or less than the minimumSendInterval) the time virt-who will wait will be the minimumSendInterval * number_of_429_responses_in_a_row. So the first 429 received for a given request (without the Retry-After header) will result in a wait time of 60s. The second will result in a wait time of 120s and so on.

### DIFF
--- a/tests/test_virt.py
+++ b/tests/test_virt.py
@@ -305,6 +305,7 @@ class TestDestinationThread(TestBase):
         # In this test we want to see that the wait method is called when we
         # expect and with what parameters we expect
         destination_thread.wait = Mock()
+        destination_thread.is_terminated = Mock(return_value=False)
         destination_thread._send_data(data_to_send)
         # We expect there two be two calls to wait with the value of the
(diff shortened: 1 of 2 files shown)


wottop:
ack

PR #68, opened 2017-05-11
1447264: Keep running on InvalidPasswordFormat given other valid configs

cnsnyder:
Seemed like a quick fix to the referenced bug.
Basically, if we fail to decrypt the password for a config, log it and continue (if there are other configs left to run with).


cnsnyder:
[Bug 1447264 - Virt-who can't be started with only one of hypervisor configs failed to be decrypted](https://trello.com/c/CQ0ceT1O/383-bug-1447264-virt-who-can-t-be-started-with-only-one-of-hypervisor-configs-failed-to-be-decrypted)

PR #69, opened 2017-05-12
1444718: Log name of config when duplicate reports are retrieved

cnsnyder:


cnsnyder:
[Bug 1444718 - It should distinguish to show "Duplicate report for config "xxx" found, ignoring" in rhsm log](https://trello.com/c/D1HMv9Eh/387-bug-1444718-it-should-distinguish-to-show-duplicate-report-for-config-xxx-found-ignoring-in-rhsm-log)

PR #70, opened 2017-05-16
1450747: Continue running destination threads on internal failure

cnsnyder:
The behaviour of virt-who prior to the implementation of the interval design, was to continue running despite connection failures when attempting to send reports.

This PR applies the same behaviour to the new design.

When reviewing this PR, in order to see the difference in behaviour for yourself, please follow the reproduction steps in the bug https://bugzilla.redhat.com/show_bug.cgi?id=1450747


wottop:
ack

PR #72, opened 2017-06-01
1457101: Continue running despite malformed configs

cnsnyder:
We'd like virt-who to continue to run when there are configurations that are invalid.
Should there be no configs that are valid virt-who will default to local libvirt.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -189,7 +189,13 @@ type=invalid
 server=1.2.3.4
 username=test
 """)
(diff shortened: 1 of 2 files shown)


cnsnyder:
[Bug 1457101 - Virt-who can't be started when only one of hypervisors' owner/env option is missing.](https://trello.com/c/xlRacPOp/428-bug-1457101---virt-who-can%27t-be-started-when-only-one-of-hypervisors%27-owner%2Fenv-option-is-missing.)

PR #73, opened 2017-06-02
1389729: virt-who incorrectly reports 'name' instead of 'hostname'

frangdlt:
1389729: virt-who incorrectly reports 'name' instead of 'hostname' for RHEV hosts.

This change makes virt-who report the actual hostname/IP address of the hypervisors, rather than the user-definable "hostname", which might not be an RFC-compliant hostname.


frangdlt:
Hi Chris -- totally missed the tests part. Now it should be ok ;-) . Thanks!

cnsnyder:
@frangdlt Thank you much! Looks good to me!

PR #74, opened 2017-06-09
1452436: virt-who prints host-to-quest mapping everytime

jirihnidek:
* Bug fix: https://bugzilla.redhat.com/show_bug.cgi?id=1452436
* When there is no hypervisor nor quest to report, then
  following message is printed to log:

      Hosts-to-guests mapping for config "sat-rhevm":
      0 hypervisors and 0 guests found

* Refactoring of code.

### DIFF
--- a/virtwho/manager/subscriptionmanager/subscriptionmanager.py
+++ b/virtwho/manager/subscriptionmanager/subscriptionmanager.py
@@ -60,6 +60,7 @@ class SubscriptionManager(Manager):
         self.cert_uuid = None
         self.rhsm_config = None
         self.readConfig()
+        self.connection = None
 
     def readConfig(self):
         """ Parse rhsm.conf in order to obtain consumer
@@ -142,6 +143,8 @@ class SubscriptionManager(Manager):
         except BadStatusLine:
             raise ManagerError("Communication with subscription manager interrupted")
 
+        return self.connection
+
     def sendVirtGuests(self, report, options=None):
         """
         Update consumer facts with info about virtual guests.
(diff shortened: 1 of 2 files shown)


jirihnidek:
[Bug 1452436 - No any Host-to-guest mapping info display to rhsm log when there isn't vdsm host on rhevm.](https://trello.com/c/yBPjPX9l/444-bug-1452436-no-any-host-to-guest-mapping-info-display-to-rhsm-log-when-there-isnt-vdsm-host-on-rhevm)

virtwho/manager/subscriptionmanager/subscriptionmanager.py
```diff
@@ -220,13 +205,64 @@ def hypervisorCheckIn(self, report, options=None):
...
+            # Reformat the data from the mapping to make it fit with
+            # the old api.
+            for hypervisor in mapping['hypervisors']:
+                guests = [g.toDict() for g in hypervisor.guestIds]
+                serialized_mapping[hypervisor.hypervisorId] = guests
+
+        hypervisor_count = len(mapping['hypervisors'])
+        guest_count = sum(len(hypervisor.guestIds) for hypervisor in mapping['hypervisors'])
+
+        self.logger.info('Hosts-to-guests mapping for config "%s": %d hypervisors and %d guests found',
```
cnsnyder: I think this log message should be output for each report gathered from each source. In other words, this should probably be output when the reports are gathered and not when they are being sent.

So if there is a configuration "libvirt_1" and a configuration "libvirt_2", I think we should see this message both times. Once for the info gathered from "libvirt_1" and another time for the info gathered from "libvirt_2".

virtwho/virt/virt.py
```diff
@@ -705,16 +709,22 @@ def _send_data(self, data_to_send):
...
-        if all((source_key in sources_sent or source_key in sources_erred)
-               for source_key in self.source_keys) and self._oneshot:
-            if not self.options.print_:
-                self.logger.debug('At least one report for each connected '
-                                  'source has been sent. Terminating.')
-            else:
-                self.logger.debug('All info to print has been gathered. '
-                                  'Terminating.')
-            self.stop()
+        if any_report_sent is False:
```
cnsnyder: As mentioned above, this condition appears to only ever be false if `data_to_send` does not include a report for all of `self.source_keys`. In this case the `batch_host_guest_report` will not have the data from all sources but will be printed once more (as it is already printed during the call to hypervisorCheckIn).

jirihnidek:
@cnsnyder Thanks for review. I will finish it in Monday.

jirihnidek:
I can confirm this PR fixes this bug too: https://bugzilla.redhat.com/show_bug.cgi?id=1445579

kahowell:
> I can confirm this PR fixes this bug too: https://bugzilla.redhat.com/show_bug.cgi?id=1445579

@jirihnidek , accordingly, please close 1445579 as a duplicate of 1452436

cnsnyder:
@jirihnidek I still believe we should have these sort of messages printed for each source before they are batched.

For example with two configurations in /etc/virt-who.d/test.conf as follows:
"""
[libvirt]
type=libvirt
server=qemu:///system
env=SOME_ENV
owner=SOME_OWNER

[libvirt_2]
type=libvirt
server=qemu:///system
env=SOME_ENV
owner=SOME_OWNER
"""

I would expect to see two log messages:

"""
2017-06-12 14:47:53,811 INFO: Hosts-to-guests mapping for config "libvirt": 1 hypervisors and 13 guests found
2017-06-12 14:48:53,811 INFO: Hosts-to-guests mapping for config "libvirt_2": 2 hypervisors and 26 guests found
"""

jirihnidek:
@cnsnyder I agree. It sounds reasonable. I'm sorry, I didn't include it in last commit.

jirihnidek:
Squashed and pushed.

PR #76, opened 2017-06-12
1357761: Do not check passwords to be in latin1 encoding

jirihnidek:
* Bug fix: https://bugzilla.redhat.com/show_bug.cgi?id=1357761
* Check only usernames to be in latin1 encoding.
* Passwords can contain unicode characters.

### DIFF
--- a/virtwho/config.py
+++ b/virtwho/config.py
@@ -373,15 +373,19 @@ class Config(GeneralConfig):
         ('filter_host_uuids', 'filter_hosts'),
         ('exclude_host_uuids', 'exclude_hosts'),
     )
+    # It is usually required to have username in latin1 encoding
     LATIN1_OPTIONS = (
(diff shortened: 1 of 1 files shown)


jirihnidek:
[Bug 1357761 - Problem with virt-who using Unicode characters in a hypervisor's account password?](https://trello.com/c/m4O8Z37Z/445-bug-1357761-problem-with-virt-who-using-unicode-characters-in-a-hypervisors-account-password)

jirihnidek:
This bug: https://bugzilla.redhat.com/show_bug.cgi?id=1437017 seems as duplicate of bug: https://bugzilla.redhat.com/show_bug.cgi?id=1357761. If I modified this PR a little, then it would be possible to address BZ 1437017 too.

jirihnidek:
I noticed after creating this PR that the BZ 1357761 is mentioned in your documentation about refactoring.

PR #77, opened 2017-06-14
1404117: Check parameter consistency and refactoring

jirihnidek:
* Bug fix:
  * https://bugzilla.redhat.com/show_bug.cgi?id=1461417
  * https://bugzilla.redhat.com/show_bug.cgi?id=1404117
* Module argparse is used now, because optparse is marked
  as obsolete and it does not provide required functionality
  to fix this bug in elegant way.
* New subclass of Action was created. It is used for
  checking consistency of command line options. When
  virtualization backend is specified using e.g. --xen,
  then --libvirt-username=admin cannot follow. There
  should be --xen-username=admin.
* Command line arguments have to be in right order. It
  means e.g. --libvirt has to be first and then
  --libvirt-username=admin can follow.
* When --libvirt-username=admin is used without --libvirt,
  then exception will be raised, because the argument
  would not have any effect.
* Refactored code a little.
* Output of virt-who --help is little bit different.

### DIFF
--- a/virtwho/parser.py
+++ b/virtwho/parser.py
@@ -1,124 +1,298 @@
+# -*- coding: utf-8 -*-
+
+# Agent for reporting virtual guest IDs to subscription-manager
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+"""
+This module is used for parsing command line arguments and reading
+configuration from environment variables.
+"""
 
 import os
-from optparse import OptionParser, OptionGroup, SUPPRESS_HELP
+from argparse import ArgumentParser, Action
 
 from virtwho import log, MinimumSendInterval, DefaultInterval
 from virtwho.config import GlobalConfig, NotSetSentinel, VIRTWHO_GENERAL_CONF_PATH
+from virtwho.virt.virt import Virt
 
 
 SAT5 = "satellite"
 SAT6 = "sam"
+
+# List of supported virtualization backends
+VIRT_BACKENDS = Virt.hypervisor_types()
 
 
 class OptionError(Exception):
     pass
 
 
-class OptionParserEpilog(OptionParser):
-    """ Epilog is new in Python 2.5, we need to support Python 2.4. """
-    def __init__(self, usage="%prog [options]", description=None, epilog=None):
-        self.myepilog = epilog
-        OptionParser.__init__(self, usage=usage, description=description)
+class StoreGroupArgument(Action):
+    """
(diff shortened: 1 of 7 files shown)


jirihnidek:
[Bug 1461417 - when running virt-who by CLI, it will not check the parameters consistency](https://trello.com/c/QLG6cPhK/446-bug-1404117-when-running-virt-who-by-cli-it-will-not-check-the-parameters-consistency)

jirihnidek:
Note: when virt-who will be executed, with reversed order of arguments. e.g.

    virt-who --xen-password=admin --xen-username=admin --libvirt

then this will leaked in. It is not problem to add another custom class.

jirihnidek:
Enable something like this:

    virt-who --xen-password=admin --xen-username=admin --xen

is not possible without huge rewrite of code. Thus virtualization backend have to be specified first:

    virt-who --xen --xen-password=admin --xen-username=admin

cnsnyder:
@jirihnidek Pending further discussion and planning from myself and the team I think it's best if we hold off on these changes for now.

jirihnidek:
I have idea. If following method call (and similar method calls):

```python
libvirt_group.add_argument(
    "--libvirt-owner",
    action=StoreGroupArgument,
    dest="owner",
    default="",
    help="Organization who has purchased subscriptions of the products, "
         "default is owner of current system"
)
```

was replaced with something like this:

```python
libvirt_group.add_argument(
    "--libvirt-owner",
    action=StoreGroupArgument,
    dest="libvirt_owner",
    default="",
    help="Organization who has purchased subscriptions of the products, "
         "default is owner of current system"
)
```

we could not require the particular order of arguments. We could save valid argument `options['libvirt_owner']` to `options['owner']` (and others) after checking all arguments.

cnsnyder:
@jirihnidek The above sounds good to me. Thank you!

jirihnidek:
Arguments can be in any order now. Added more tests. Pushed and squashed :-).

I vote for removing all group arguments for setting options of virtual backends in the future.

virtwho/parser.py
```diff
@@ -9,116 +31,267 @@
 SAT5 = "satellite"
 SAT6 = "sam"
 
+# List of supported virtualization backends
+VIRT_BACKENDS = ['libvirt', 'vdsm', 'esx', 'rhevm', 'hyperv', 'xen']
```
cnsnyder: Might I suggest that this list be defined (or rather computed) via something in the Virt base class? (virtwho/virt/virt.py)
As each of these values exist as a class attribute on each subclass of Virt, and Virt already checks through all of these values in the "from_config" method, we might want to avoid having to update this list in two places should there be any additions or removals in this area

virtwho/parser.py
```diff
@@ -9,116 +31,267 @@
...
+    )
+    manager_group.add_argument("--sam", action="store_const", dest="smType", const=SAT6, default=SAT6,
+                               help="Report host/guest associations to the Subscription Asset Manager [default]")
+    manager_group.add_argument("--satellite6", action="store_const", dest="smType", const=SAT6,
+                               help="Report host/guest associations to the Satellite 6 server")
+    manager_group.add_argument("--satellite5", action="store_const", dest="smType", const=SAT5,
+                               help="Report host/guest associations to the Satellite 5 server")
+    manager_group.add_argument("--satellite", action="store_const", dest="smType", const=SAT5)
+
+    # FIXME: Remove all options of virtualization backend. Adding this wasn't happy design decision.
```
cnsnyder: Agreed, I think we should do this sooner rather than later. (Non-nacking comment)

virtwho/parser.py
jirihnidek: I agree, very good point. Thanks for review.

jirihnidek:
Commited, squashed and force pushed.

PR #78, opened 2017-06-19
1401420: xen supports only uuid/hostname as hypervisor_id

jirihnidek:
* Bug fix: https://bugzilla.redhat.com/show_bug.cgi?id=1401420
* Manual pages already contains this information:
  "hwuuid is applicable to esx and rhevm only."
* This bug is caused probably by copy pasting code from esx/rhevm.

### DIFF
--- a/virtwho/virt/xen/xen.py
+++ b/virtwho/virt/xen/xen.py
@@ -123,13 +123,11 @@ class Xen(virt.Virt):
 
             if self.config.hypervisor_id == 'uuid':
                 uuid = record["uuid"]
-            elif self.config.hypervisor_id == 'hwuuid':
-                uuid = record["cpu_info"]['features']
             elif self.config.hypervisor_id == 'hostname':
                 uuid = record["hostname"]
             else:
                 raise virt.VirtError(
-                    'Invalid option %s for hypervisor_id, use one of: uuid, hwuuid, or hostname' %
+                    'Invalid option %s for hypervisor_id, use one of: uuid or hostname' %
                     self.config.hypervisor_id)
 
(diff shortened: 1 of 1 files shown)


jirihnidek:
[Bug 1415064 - [Virt-who] hypervisor_id config can't work when virt-who run at xen mode](https://trello.com/c/i4VuMbu6/465-bug-1415064-virt-who-hypervisorid-config-cant-work-when-virt-who-run-at-xen-mode)

cnsnyder:
@jirihnidek I wonder if there is something else that could be used for the hwuuid, something else that would be more unique. I would also be curious to know if the same issue exists already for esx and / or rhevm?

cnsnyder:
@jirihnidek Looking at the definition of hwuuid for esx, it seems xen and esx are using quite different values for hwuuid. So I think removing it in the meantime is ok. We might want to consider finding an alternative in the future though.

jirihnidek:
@cnsnyder I doubt we can find anything, where we can guarantee uniqueness (Xen, libvirtd). In theory it could be hypervisor MAC address of physical network interface, but in practice neither MAC address it is not 100% unique. When I look at json report from xen system, I don't see any MAC address there though.

I think it is OK to allow `hwuuid` only for esx and rhevm, because these systems are masters of `hwuuid` uniqueness. Thus this uniqueness can be guaranteed.

vritant:
[[13] [performace tests] Look at existing test failures and adjust expected success criteria](https://trello.com/c/OElgNmQs/412-13-performace-tests-look-at-existing-test-failures-and-adjust-expected-success-criteria)

PR #79, opened 2017-06-23
1458184: Better reading of environment variables

jirihnidek:
* Bug fix: https://bugzilla.redhat.com/show_bug.cgi?id=1458184
* Huge refactoring of code

### DIFF
--- a/virtwho/parser.py
+++ b/virtwho/parser.py
@@ -35,6 +35,22 @@ SAT6 = "sam"
 # List of supported virtualization backends
 VIRT_BACKENDS = Virt.hypervisor_types()
 
+SAT_VM_DISPATCHER = {
+    'libvirt': {'owner': False, 'env': False, 'server': False, 'username': False},
+    'esx': {'owner': False, 'env': False, 'server': True, 'username': True},
+    'xen': {'owner': False, 'env': False, 'server': True, 'username': True},
+    'rhevm': {'owner': False, 'env': False, 'server': True, 'username': True},
+    'hyperv': {'owner': False, 'env': False, 'server': True, 'username': True},
+}
+
+SAM_VM_DISPATCHER = {
+    'libvirt': {'owner': False, 'env': False, 'server': False, 'username': False},
+    'esx': {'owner': True, 'env': True, 'server': True, 'username': True},
+    'xen': {'owner': True, 'env': True, 'server': True, 'username': True},
+    'rhevm': {'owner': True, 'env': True, 'server': True, 'username': True},
+    'hyperv': {'owner': True, 'env': True, 'server': True, 'username': True},
+}
+
 
 class OptionError(Exception):
     pass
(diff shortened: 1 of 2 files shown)


jirihnidek:
[Bug 1458184 - Virt-who can't be started when username option is none/null under /etc/sysconfig/virt-who](https://trello.com/c/61U256E4/450-bug-1458184-virt-who-cant-be-started-when-username-option-is-none-null-under-etc-sysconfig-virt-who)

virtwho/parser.py
```diff
@@ -112,21 +120,151 @@ def check_argument_consistency(cli_options):
     Final check of cli options that can not be done in custom actions.
     """
 
+    # These options can be required
+    REQUIRED_OPTIONS = ['owner', 'env', 'server', 'username']
```
jirihnidek: This part of refactoring was done, because current implementation uses one function for checking consistency of environment variables ad command line arguments. Thus error logs can be confusing.

virtwho/parser.py
```diff
@@ -35,6 +35,14 @@
 # List of supported virtualization backends
 VIRT_BACKENDS = Virt.hypervisor_types()
 
+VM_DISPATCHER = {
```
jirihnidek: I agree that this type of validation should be part of class.

jirihnidek:
I'm almost sure that this PR fixes bug from bug report, but I'm sure it does not solve all possible combinations of wrong and valid values of environment variables, configuration files and command line arguments.

Environment variables and sysconfig (and command line arguments) should be IMHO used only for basic settings like `debug`. Most of things should be possible to do only in configuration files (I know ... backward compatibility). Priorities between env. variables and corresponding command line arguments should be covered by set of unit tests. I use my intuition what is right behavior many times and it is not right approach.

I feel I'm in strong interference with your refactoring effort in this PR. What should I do? Close this PR?

cnsnyder:
If I've understood the bug correctly, the following case (and similar) should still run:
1) Set some environment variables that specify a virt backend (one or more of which should be bad values)

`export VIRTWHO_HYPERV=1
export VIRTWHO_HYPERV_SERVER=1
export VIRTWHO_HYPERV_USERNAME=1`

2) Set up some valid configuration section in a file in the drop dir (/etc/virt-who.d/*.conf)

"""
[test_libvirt]
type=libvirt
server=qemu:///system
env=admin
owner=admin
"""

3) Run virt-who


Expected results:
Virt-who runs and logs a message showing that the env/cli backend is somehow invalid. Virt-who continues to run with the valid configurations.


Actual Results:
Virt-who logs a message about the invalid nature of the Env vars and CLI but immediately quits running.

jirihnidek:
Yeah, `VIRTWHO_ESX_ENV` and `VIRTWHO_ESX_OWNER` were handled by old code. I did more refactoring. I'm crossing fingers and it should be fine for your example too now. :-)

PR #81, opened 2017-07-05
Add VWEffectiveConfig

cnsnyder:
This PR adds a config object that will largely replace much of the remainder of config.py.
This EffectiveConfig (or VWEffectiveConfig as it is currently named), represents the total configuration that virt-who will use while running.
The new config object takes in the cli_options (parsed from the command line) and the env_options (parsed from the environement) as well as the contents of /etc/virt-who.conf as well as /etc/virt-who.d/*.conf.

This object implements the hierarchy from which the effective values for a given configuration section of virt-who are derived. In addition this object validates the global values (those values that affect virt-who as a whole).

This PR is not meant for master yet as it is the first step towards (hopefully) cleaner configuration in virt-who. That being said the tests should pass with this PR. The VWEffectiveConfig has been integrated into virt-who and is now used in place of the GlobalConfig.

Please note: I would like to squash these commits down once review (and requisite clean up) has taken place.

Also note: config.EnvParser is an alternative implementation of a parser of environment variables. It can likely be removed but I could see an argument for using that version instead.

PR #82, opened 2017-07-10
An alternative EffectiveConfig

cnsnyder:
This PR adds the following classes:
- EffectiveConfig: A container object that represents the effective configuration that virt-who is currently running with. It maintains the hierarchy of configuration sources applied to all sections. It has dictionary like access. It contains ConfigSection objects.

- ConfigSection: This is the base class for each type of section we might have in virt-who configuration. It contains the logic to validate a section of the configuration as well as serves as a collection of that configuration. It also has dictionary like access. It also has a state variable which it updates every time the validate method is run. The validate method returns a list of ("log_level", "message"), which are meant to be any sort of validation output that the section deems necessary. It is also the responsibility of the validate method to deserialize (if necessary) any of the parameters given it (so that anything using that ConfigSection in the code base need not worry about that).

- GlobalConfigSection: A subclass of ConfigSection. Implements the validation method and fills in some DEFAULTS. Note not all configuration sections will be validated the same way so this might not be the very best example of how to write a ConfigSection for a virt backend. it should be fairly similar though.

The remainder of the changes were to get the EffectiveConfig in at least some use throughout the code base. The EffectiveConfig has essentially replaced the "options" objects seen throughout the code base. Other updates were to mainly to fix broken tests.

### DIFF
--- a/virtwho/config.py
+++ b/virtwho/config.py
@@ -18,21 +18,31 @@ along with this program; if not, write to the Free Software
 Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 """
 
+import collections
 import os
-
 from ConfigParser import SafeConfigParser, NoOptionError, Error, MissingSectionHeaderError
-from virtwho import DefaultInterval
+from virtwho import log
 from password import Password
 from binascii import unhexlify
 import hashlib
 import json
 import util
 
-VIRTWHO_CONF_DIR = "/etc/virt-who.d/"
-VIRTWHO_TYPES = ("libvirt", "vdsm", "esx", "rhevm", "hyperv", "fake", "xen")
-VIRTWHO_GENERAL_CONF_PATH = "/etc/virt-who.conf"
-VIRTWHO_GLOBAL_SECTION_NAME = "global"
-VIRTWHO_VIRT_DEFAULTS_SECTION_NAME = "defaults"
+# Module-level logger
+logger = log.getLogger('config', queue=False)
+
+_effective_config = None
+
+VW_CONF_DIR = "/etc/virt-who.d/"
+VW_TYPES = ("libvirt", "vdsm", "esx", "rhevm", "hyperv", "fake", "xen")
+VW_GENERAL_CONF_PATH = "/etc/virt-who.conf"
+VW_GLOBAL = "global"
+VW_VIRT_DEFAULTS_SECTION_NAME = "defaults"
+VW_ENV_CLI_SECTION_NAME = "env/cmdline"
+
+# Default interval for sending list of UUIDs
+DefaultInterval = 3600  # One per hour
+MinimumSendInterval = 60  # One minute
 
 
 class InvalidOption(Error):
@@ -299,7 +309,7 @@ class GeneralConfig(object):
         return True
 
     @classmethod
-    def fromFile(cls, filename, logger):
+    def from_file(cls, filename):
         raise NotImplementedError()
 
 
@@ -312,7 +322,7 @@ class GlobalConfig(GeneralConfig):
     DEFAULTS = {
         'debug': False,
         'oneshot': False,
-        'print_': False,
+        'print': False,
         'log_per_config': False,
         'background': False,
         'configs': '',
@@ -327,7 +337,7 @@ class GlobalConfig(GeneralConfig):
         'debug',
         'oneshot',
         'background',
-        'print_'
+        'print'
         'log_per_config'
     )
     INT_OPTIONS = (
@@ -335,13 +345,13 @@ class GlobalConfig(GeneralConfig):
     )
 
     @classmethod
-    def fromFile(cls, filename, logger=None):
-        global_config = parseFile(filename, logger=logger).get(VIRTWHO_GLOBAL_SECTION_NAME)
+    def from_file(cls, filename):
+        global_config = parse_file(filename).get(VW_GLOBAL)
         if not global_config:
             if logger:
                 logger.warning(
                     'Unable to find "%s" section in general config file: "%s"\nWill use defaults where required',
-                    VIRTWHO_GLOBAL_SECTION_NAME, filename)
+                    VW_GLOBAL, filename)
             global_config = {}
         return cls(**global_config)
 
@@ -387,9 +397,9 @@ class Config(GeneralConfig):
         self._name = name
         self._type = virtwho_type
 
-        if self._type not in VIRTWHO_TYPES:
+        if self._type not in VW_TYPES:
             raise InvalidOption('Invalid type "%s", must be one of following %s' %
-                                (self._type, ", ".join(VIRTWHO_TYPES)))
+                                (self._type, ", ".join(VW_TYPES)))
 
         for password_option, decrypted_option in self.PASSWORD_OPTIONS:
             try:
@@ -523,14 +533,14 @@ class ConfigManager(object):
     def __init__(self, logger, config_dir=None, defaults=None):
         if not defaults:
             try:
-                defaults_from_config = parseFile(VIRTWHO_GENERAL_CONF_PATH).get(VIRTWHO_VIRT_DEFAULTS_SECTION_NAME)
+                defaults_from_config = parse_file(VW_GENERAL_CONF_PATH).get(VW_VIRT_DEFAULTS_SECTION_NAME)
                 self._defaults = defaults_from_config or {}
             except MissingSectionHeaderError:
                 self._defaults = {}
         else:
             self._defaults = defaults
         if config_dir is None:
-            config_dir = VIRTWHO_CONF_DIR
+            config_dir = VW_CONF_DIR
         parser = StripQuotesConfigParser()
         self._configs = []
         self.logger = logger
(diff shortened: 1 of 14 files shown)


cnsnyder:
Reviewer: Please merge only #81 OR #82

cnsnyder:
[[13] [CONFIG_EPIC] Create ConfigHierarchy Class](https://trello.com/c/RF32mpFs/451-13-configepic-create-confighierarchy-class)

virtwho/config.py
```diff
@@ -18,21 +18,30 @@
...
 import os
-
 from ConfigParser import SafeConfigParser, NoOptionError, Error, MissingSectionHeaderError
-from virtwho import DefaultInterval
+from virtwho import log
 from password import Password
 from binascii import unhexlify
 import hashlib
 import json
+import six
```
jirihnidek: Module `six` is not used. I would not mix this refactoring with effort to add support for Python 3.

virtwho/config.py
```diff
@@ -18,21 +18,30 @@
...
 
-VIRTWHO_CONF_DIR = "/etc/virt-who.d/"
+# Module-level logger
+logger = log.getLogger('config', queue=False)
+
+VW_CONF_DIR = "/etc/virt-who.d/"
 VIRTWHO_TYPES = ("libvirt", "vdsm", "esx", "rhevm", "hyperv", "fake", "xen")
 VIRTWHO_GENERAL_CONF_PATH = "/etc/virt-who.conf"
-VIRTWHO_GLOBAL_SECTION_NAME = "global"
+VW_GLOBAL = "global"
```
jirihnidek: `VW` stands for Volkswagen, right? ;-) ... Seriously, I would use VW in other constants too.

virtwho/config.py
```diff
@@ -327,21 +336,21 @@ class GlobalConfig(GeneralConfig):
...
         'log_per_config'
     )
     INT_OPTIONS = (
         'interval',
     )
 
     @classmethod
-    def fromFile(cls, filename, logger=None):
-        global_config = parseFile(filename, logger=logger).get(VIRTWHO_GLOBAL_SECTION_NAME)
+    def fromFile(cls, filename):
```
jirihnidek: I would not use camelCase and I would rename it to: `from_file()`.

virtwho/config.py
```diff
@@ -685,29 +694,487 @@ def addConfig(self, config):
...
+    for section in sections:
+        if '__name__' in sections[section]:
+            del sections[section]['__name__']
     return sections
+
+# Helper methods used to validate parameters given to virt-who
+def str_to_bool(value):
+    if isinstance(value, bool):
+        return value
+    if isinstance(value, str):
```
jirihnidek: To advocate importing `six` module you can use here: `if isinstance(value, six.string_types):

virtwho/config.py
```diff
@@ -685,29 +694,487 @@ def addConfig(self, config):
...
-    for section in parser.sections():
-        try:
-            sections[section] = getOptions(section, parser)
-        except NoOptionError:
-            sections[section] = {}
-    return sections
-
-
-def parseFile(filename, logger=None):
+def parseFile(filename):
```
jirihnidek: I would use: `def parse_file(filename)`

virtwho/config.py
```diff
@@ -299,7 +308,7 @@ def has_options(self, options):
         return True
 
     @classmethod
-    def fromFile(cls, filename, logger):
+    def fromFile(cls, filename):
```
jirihnidek: `from_file()`

virtwho/config.py
```diff
@@ -685,29 +694,487 @@ def addConfig(self, config):
...
+        raise TypeError("Value is not a string")
+    if not value:
+        raise ValueError("String is empty")
+    return value
+
+
+def readable(path):
+    if not os.access(path, os.R_OK):
+        raise ValueError("Path '%s' is not accessible" % path)
+    return path
```
jirihnidek: When you return value, then it suggests that you do something with `path`. It is not this case.

virtwho/config.py
```diff
@@ -685,29 +694,487 @@ def addConfig(self, config):
...
+
+
+def accessible_file(path):
+    readable(path)
+    if not os.path.isfile(path):
+        raise ValueError("Path '%s' does not appear to be a file" % path)
+    return path
```
jirihnidek: Same here.

virtwho/config.py
```diff
@@ -685,29 +694,487 @@ def addConfig(self, config):
...
+    if not os.path.isdir(path):
+        raise ValueError("Path '%s' does not appear to be a directory" % path)
+
+
+def empty_or_accessible_files(paths):
+    if not isinstance(paths, list):
+        if not isinstance(paths, str):
+            raise TypeError()
+        if len(paths) == 0:
+            return []
```
jirihnidek: I guess this case is solved at line 754 too.

virtwho/config.py
```diff
@@ -685,29 +694,487 @@ def addConfig(self, config):
...
+def accessible_dir(path):
+    readable(path)
+    if not os.path.isdir(path):
+        raise ValueError("Path '%s' does not appear to be a directory" % path)
+
+
+def empty_or_accessible_files(paths):
+    if not isinstance(paths, list):
+        if not isinstance(paths, str):
+            raise TypeError()
```
jirihnidek: What about adding some string to this `TypeError()`. Something like: `TypeError('Path has to be list of strings or string')

virtwho/config.py
```diff
@@ -685,29 +694,487 @@ def addConfig(self, config):
...
+    return path
+
+
+def accessible_dir(path):
+    readable(path)
+    if not os.path.isdir(path):
+        raise ValueError("Path '%s' does not appear to be a directory" % path)
+
+
+def empty_or_accessible_files(paths):
```
jirihnidek: I would add some doc string for this method, because the name is no self-explanatory.

virtwho/config.py
```diff
@@ -685,29 +694,487 @@ def addConfig(self, config):
...
+    return paths
+
+
+class ConfigSection(collections.MutableMapping):
+    """
+    This represents a section of configuration for virt-who. The interface that it exposes is 
+    dictionary like. This object maintains a state attribute. The state shows if the configuration
+    section has passed validation, needs validation, or is invalid for some reason.
+    """
+    # The string representation of all default properties and values
```
jirihnidek: Add note that real values will be added in child classes.

virtwho/config.py
```diff
@@ -685,29 +694,487 @@ def addConfig(self, config):
...
+    DEFAULTS = ()
+    REQUIRED = ()
+
+    __marker = object()
+
+    def __init__(self, section_name, wrapper):
+        self.section_name = section_name
+        self._wrapper = wrapper
+        self.defaults = dict(type(self).DEFAULTS)
```
jirihnidek: Why not just `self.DEFAULTS`?

virtwho/config.py
```diff
@@ -685,29 +694,487 @@ def addConfig(self, config):
...
+        # Those properties that have yet to be validated
+        # Any item not in this set but in this ConfigSection is validated
+        self._unvalidated_keys = set()
+        self._invalid_keys = set()
+        # Holds all properties for this section
+        self._values = {}
+        # Add section defaults
+        for key, value in self.defaults.iteritems():
```
jirihnidek: Not Python 3 friendly, use `self.defaults.items()`.

virtwho/config.py
```diff
@@ -685,29 +694,487 @@ def addConfig(self, config):
...
+                for key, value in arg:
+                    self._values[key] = value
+        for key, value in kwds.iteritems():
+            self._values[key] = value
+
+    @classmethod
+    def get_defaults(cls):
+        defaults = dict()
+        for key, value in cls.DEFAULTS:
+            defaults[key] = str(value)
```
jirihnidek: Why casting to `str`?

virtwho/config.py
```diff
@@ -685,29 +694,487 @@ def addConfig(self, config):
...
+
+    def update(self, *args, **kwds):
+        for arg in args:
+            if getattr(arg, 'keys', None):
+                for key in arg.keys():
+                    self._values[key] = arg[key]
+            else:
+                for key, value in arg:
+                    self._values[key] = value
+        for key, value in kwds.iteritems():
```
jirihnidek: Use `kwds.items()` to be compatible with Python 3.

virtwho/config.py
```diff
@@ -685,29 +694,487 @@ def addConfig(self, config):
...
+                return default
+            raise e
+
+    def __str__(self):
+        return_string = "[%s]" % self.section_name
+        for key, value in self.iteritems():
+            return_string += "\n%s=%s" % (key, value)
+        return return_string
+
+    def update(self, *args, **kwds):
```
jirihnidek: This method deserves doc string.

virtwho/config.py
```diff
@@ -685,29 +694,487 @@ def addConfig(self, config):
...
+        return self.defaults[key] == self._values[key]
+
+    def has_default(self, key):
+        return key in self.defaults
+
+    def get(self, key, default=__marker):
+        try:
+            return self[key]
+        except KeyError as e:
+            if default is not self.__marker:
```
jirihnidek: What? Why? In what case this is not `True`?

virtwho/config.py
```diff
@@ -685,29 +694,487 @@ def addConfig(self, config):
...
+            defaults[key] = str(value)
+        return defaults
+
+    @classmethod
+    def class_for_type(cls, virt_type):
+        clazz = ConfigSection
+        # TODO: Update to use actual subclasses as the are created
+        for subclass in cls.__subclasses__():
+            if subclass.VIRT_TYPE == virt_type:
+                clazz = subclass
```
jirihnidek: I would add `break` here.

virtwho/config.py
```diff
@@ -685,29 +694,487 @@ def addConfig(self, config):
...
+                                                   'See man virt-who-config for more info')
+        return result
+
+    def _validate_configs(self):
+        result = None
+        if self.is_default('configs'):
+            self._values['configs'] = []
+        elif isinstance(self['configs'], list):
+            self._values['configs'] = [item for item in self._values['configs'] if isinstance(item,
+                                                                                          str)]
```
jirihnidek: I hope it is possible to break the line in better way.

virtwho/config.py
```diff
@@ -685,29 +694,487 @@ def addConfig(self, config):
...
+        elif not isinstance(self._values['configs'], str):
+            result = ('warning', '"configs" must be one or more strings, '
+                                                   'ignoring')
+            self._values['configs'] = []  # Reset to empty list
+        else:
+            self._values['configs'] = self._values['configs'].split(',')
```
jirihnidek: I would switch elif-else here to something like this:

```python
elif isinstance(self._values['configs'], str):
    # ...
else:
    # ...
```

virtwho/config.py
```diff
@@ -685,29 +694,487 @@ def addConfig(self, config):
...
+            result = ('warning', '%s cannot be empty, using default' % key)
+        return result
+
+    def validate(self, validation_messages=None):
+        validation_messages = super(GlobalSection, self).validate(validation_messages)
+        to_reset = set()
+        # Validate those keys that need to be validated
+        for key in set(self._unvalidated_keys):
+            error = None
+            # If boolean
```
jirihnidek: ?

virtwho/config.py
```diff
@@ -685,29 +694,487 @@ def addConfig(self, config):
...
+            elif key == 'configs':
+                error = self._validate_configs()
+            elif key in ['reporter_id', 'log_file', 'log_dir']:
+                error = self._validate_non_empty_string(key)
+            elif key == 'interval':
+                error = self._validate_interval()
+            else:
+                validation_messages.append(('warning', 'Ignoring unknown configuration option '
+                                                       '"%s"' % key))
+                del self._values[key]
```
jirihnidek: I would use `dict` dispatcher in this case.

virtwho/config.py
```diff
@@ -685,29 +694,487 @@ def addConfig(self, config):
...
+            new_section.update(**virt_defaults_section)
+            new_section.update(**values)
+            try:
+                new_section = ConfigSection.from_dict(new_section, section, self)
+            except KeyError:
+                # Missing required attribute
+                pass
+            self._sections[section] = new_section
+
+
```
jirihnidek: Too much white space.

virtwho/config.py
cnsnyder: To allow the default to be None I had to set the optional arg to something other than None.
This way if the user sets default to anything (so long as it is not self.__marker) and the given key is not in self, we'll return default.

virtwho/config.py
```diff
@@ -685,29 +694,487 @@ def addConfig(self, config):
...
+                error = self._validate_interval()
+            else:
+                validation_messages.append(('warning', 'Ignoring unknown configuration option '
+                                                       '"%s"' % key))
+                del self._values[key]
+            if error is not None:
+                validation_messages.append(error)
+                if key not in ['interval']:  # Special cases not reset to default on failure
+                    to_reset.add(key)
+                self._invalid_keys.add(key)
```
cnsnyder: Yea I could see that I'll figure out another name. The intention behind it is a set of keys whose values are invalid.

virtwho/config.py
```diff
@@ -685,29 +694,487 @@ def addConfig(self, config):
...
+    def __len__(self):
+        return len(self._sections)
+
+    def __getitem__(self, key):
+        return self._sections[key]
+
+    def __contains__(self, item):
+        return item in self._sections
+
+    def __init__(self, env_args, cli_args):
```
cnsnyder: Would the preference to be have a module level method which creates an instance of the EffectiveConfig object?

virtwho/config.py
cnsnyder: Let's comeback to python 3 compatibility once we've looked at the changes necessary.
For now I'll go with the removal of six. (which I added accidentally).

virtwho/config.py
cnsnyder: Sounds good, done!

PR #83, opened 2017-07-11
1400788: Do not try to get host-to-guest mapping on unregistered system

jirihnidek:
* Bug fix: https://bugzilla.redhat.com/show_bug.cgi?id=1400788
* When destinatin thread is not able to send data to candlepin server,
  then it will not try to gather host-to-guest mapping during next
  time interval.
* Added method: `is_registered()` to class `IntervalThread`.
  The name of this method could be more accurate or implementation
  could be more completed, but it would require quite big refactoring
  of code.
* Fixed several unit tests to use Datastore and not Queue.


virtwho/virt/virt.py
```diff
@@ -637,6 +657,7 @@ def _send_data(self, data_to_send):
                     break
                 self.wait(wait_time=self.interval_modifier)
                 self.interval_modifier = 0
+                self.config.is_registered = False
```
cnsnyder: Where else is this config value used? Is it needed?

virtwho/virt/virt.py
```diff
@@ -301,23 +301,42 @@ def __init__(self, logger, config, source=None, dest=None,
...
 
+    def is_registered(self):
+        """
+        This function tries to guess if system is registered according last attempt of
+        sending data to candlepin server.
+        :return: Return true, when virt-who was able to send data to candlepin server.
+        """
+
+        try:
+            is_registered = self.dest.get(key='is_registered', default=True)
```
cnsnyder: Using the existing datastore that is intended for reports allows for a sort of strange case to appear.
If a user has defined a configuration section with the name '[is_registered]' that value will end up being the value used here.

I've done this just to see what the behaviour is. For me the result is the destination thread refuses to send anything.

The config I used was as follows:

```ini
[is_registered]
type=libvirt
server=qemu:///system
owner=admin
env=admin
rhsm_hostname=http://localhost
rhsm_port=8443
rhsm_username=admin
rhsm_password=admin
rhsm_prefix=/candlepin
```

I think that storing system state in a shared (locked) dictionary is a good idea but perhaps it should be a different dictionary. Perhaps restrict the user from naming sections whatever values we'd like to store in the shared dict?

The other issue that these changes causes is when there are multiple destination threads (less common but expected). When that is the case in theory they'd all be trying to set the same value 'is_registered'.

virtwho/virt/xen/xen.py
```diff
@@ -187,8 +187,9 @@ def _run(self):
                 events = []
 
             if initial or len(events) > 0 or delta > 0:
-                assoc = self.getHostGuestMapping()
-                self._send_data(virt.HostGuestAssociationReport(self.config, assoc))
+                if self.is_registered() is True:
```
cnsnyder: I'm not entirely certain that this will solve the issue. I'm guessing the interrupted sys call happens due to this thread being killed while active waiting.

virtwho/virt/virt.py
jirihnidek: No, this is artifact of some experiment. I will delete it.

cnsnyder:
[Bug 1400788 [virt-who] - It will show "Interrupted system call" when unregister system](https://trello.com/c/UTPlo7zj/479-bug-1400788-virt-who-it-will-show-interrupted-system-call-when-unregister-system)

virtwho/virt/xen/xen.py
jirihnidek: I will do more experiments with this, but current implementation was improvement.

virtwho/virt/virt.py
jirihnidek: I didn't know that there could be more destination threads (for more candlepin servers?)

I will try to use different datastore.

virtwho/virt/xen/xen.py
jirihnidek: When I'm looking at times in this log: https://bugzilla.redhat.com/attachment.cgi?id=1228247 ('Interrupted system call': 2016-12-02 00:18:21 and 'Unable to read certificate, system is not registered or you are not root': 2016-12-02 00:18:27) I guess the interrupted system call can be just some coincidence. And further consider this: virt-who at the first time tries to collect host-to-guest mapping and then it tries to send these results to candlepin.

jirihnidek:
I improved the PR. Hopefully I fixed all changes you requested.

virtwho/virt/virt.py
```diff
@@ -714,6 +752,18 @@ def _send_data(self, data_to_send):
                         if self._oneshot:
                             sources_erred.append(source_key)
                         retry = False  # Only retry on 429
+                        consumers = self.shared_data.get('consumers', [])
```
cnsnyder: Since all destination threads and source threads have access to the same shared_data, I have a feeling that changes to this could be lost. For example, a second destination thread might be in the else cause below this comment, whichever thread puts it's list in last wins.

jirihnidek:
@cnsnyder You are probably right about it. I will try to reproduce reporting of host-to-guest mapping to two organizations. I tried to reproduce it with one Katello installation, but I wasn't successful. Will I need to use ESX for this purpose?

cnsnyder:
@jirihnidek You will likely need to use ESX to test the functionality of the ESX backends.
You could also test this against any of the other backends so modified.
Reporting to one good candlepin and another fake or bad one. That should expose the issue.
If you need any help getting virt-who configured to report to multiple destinations or set up in this way, please let me know.

cnsnyder:
This is good work, but I'm going to have to close it for now. See trello card for more details.

PR #85, opened 2017-07-19
Point Source0 to GitHub

cnsnyder:
This updates the spec file Source to point to GitHub released tar.gz files.

### DIFF
--- a/virt-who.spec
+++ b/virt-who.spec
@@ -4,15 +4,20 @@
 %global python2_sitelib %{python_sitelib}
 %endif
 
+%global release_number 1
+
+%global git_tag %{name}-%{version}-%{release_number}
+
+
 Name:           virt-who
 Version:        0.20.3
-Release:        1%{?dist}
+Release:        %{release_number}%{?dist}
 Summary:        Agent for reporting virtual guest IDs to subscription-manager
 
 Group:          System Environment/Base
 License:        GPLv2+
 URL:            https://github.com/virt-who/virt-who
-Source0:        https://fedorahosted.org/releases/v/i/virt-who/%{name}-%{version}.tar.gz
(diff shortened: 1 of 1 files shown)


evgeni:
I'd argue that having `%{release}` in the tarball name is wrong, as `%{release}` contains also the disttag, and here I end up with `virt-who-0.20.2-1.fc26.tar.gz`.

evgeni:
I think we can just hard-code `-1` here, as we only care for the main source, which should not change between revisions (the spec will, but that's outside the tarball).

cnsnyder:
@evgeni By hard code, do you mean to update the Source0 whenever we produce a new build?

evgeni:
No, I mean setting:
```
Source0:        https://codeload.github.com/%{name}/%{name}/tar.gz/%{name}-%{version}-1#/%{name}-%{version}.tar.gz
```
This will:
* download the code from codeload.g.c (thus saving a redirect which some downloaders don't like)
* name the file `%{name}-%{version}.tar.gz` without any `-%{release}` in it
* always fetch the tarball of the first dist-revision (`-1`) of a release

I think that unconditionally getting the `-1` revision for `Source0` is OK, as we only care for the main `virt-who` code here. Any further revision would maybe change the spec file or add patches (as further sources) but should never touch the main source.

evgeni:
Or we create a macro `%{release_number}` and use that like pulp does: https://github.com/pulp/pulp/blob/master/pulp.spec#L49

cnsnyder:
Updated based on feedback to mimic the way pulp handles release_number

evgeni:
Can you add `#/%{name}-%{version}.tar.gz` to the `Source0` URL? That will allow the downloader to rename the file properly, otherwise it would be called `%{git_tag}`.

evgeni:
:+1: LGTM!

cnsnyder:
Merging, test failures are errors not failures and not due to these changes.

PR #87, opened 2017-07-20
Begin use of EffectiveConfig Sections in creation of virt backends

cnsnyder:
This PR finishes laying the ground work for using the new ConfigSections to create virt backends.

Please note: As there are no backend specific ConfigSections yet (especially including any that handle encrypted password bits) there are a number of tests which do not pass and will not pass.

I fully expect all tests to pass when the config_refactor branch is merged to master but for this step I think it's necessary for them not to (in order to keep the PRs at least somewhat small).

For testing purposes I expect the libvirt backend to work as expected in all forms of configuration. I've added a temporary commit to modify the libvirt backend to work with a generic config section. Encrypted passwords will still not work with the libvirt backend.

### DIFF
--- a/virtwho/config.py
+++ b/virtwho/config.py
@@ -504,6 +504,11 @@ class Config(GeneralConfig):
         config = Config(name, virtwho_type, defaults, **options)
         return config
 
+    @classmethod
+    def from_config_section(cls, name, section):
+        virtwho_type = section.get('type')
+        return Config(name, virtwho_type, **section)
+
     @property
     def hash(self):
         return hashlib.sha256(json.dumps(self.__dict__, sort_keys=True)).hexdigest()
@@ -529,60 +534,19 @@ class StripQuotesConfigParser(SafeConfigParser):
         return value
 
 
-class ConfigManager(object):
-    def __init__(self, logger, config_dir=None, defaults=None):
-        if not defaults:
-            try:
-                defaults_from_config = parse_file(VW_GENERAL_CONF_PATH).get(VW_VIRT_DEFAULTS_SECTION_NAME)
-                self._defaults = defaults_from_config or {}
-            except MissingSectionHeaderError:
-                self._defaults = {}
-        else:
-            self._defaults = defaults
-        if config_dir is None:
-            config_dir = VW_CONF_DIR
-        parser = StripQuotesConfigParser()
-        self._configs = []
+class DestinationToSourceMapper(object):
+    def __init__(self, effective_config):
+        self._configs = effective_config.virt_sections()
         self.logger = logger
         self.sources = set()
         self.dests = set()
         self.dest_to_sources_map = {}
-        all_dir_content = None
-        conf_files = None
-        non_conf_files = None
-        try:
-            all_dir_content = set(os.listdir(config_dir))
-            conf_files = set(s for s in all_dir_content if s.endswith('.conf'))
-            non_conf_files = all_dir_content - conf_files
-        except OSError:
-            self.logger.warn("Configuration directory '%s' doesn't exist or is not accessible", config_dir)
-            return
-        if not all_dir_content:
-            self.logger.warn("Configuration directory '%s' appears empty", config_dir)
-        elif not conf_files:
-            self.logger.warn("Configuration directory '%s' does not have any '*.conf' files but "
-                             "is not empty", config_dir)
-        elif non_conf_files:
-            self.logger.debug("There are files in '%s' not ending in '*.conf' is this "
-                              "intentional?", config_dir)
-
-        for conf in conf_files:
-            if conf.startswith('.'):
-                continue
-            try:
-                filename = parser.read(os.path.join(config_dir, conf))
-                if len(filename) == 0:
-                    self.logger.error("Unable to read configuration file %s", conf)
-            except MissingSectionHeaderError:
-                self.logger.error("Configuration file %s contains no section headers", conf)
-
-        self._readConfig(parser)
+        #self._read_effective_config(effective_config=effective_config)
         self.update_dest_to_source_map()
 
     def update_dest_to_source_map(self):
         sources, dests, d_to_s, orphan_sources = \
-            ConfigManager.map_destinations_to_sources(
-                self._configs)
+            DestinationToSourceMapper.map_destinations_to_sources(self._configs)
         if orphan_sources:
             d_to_s[default_destination_info] = orphan_sources
             dests.add(default_destination_info)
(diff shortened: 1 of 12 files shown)


cnsnyder:
[[8] [CONFIG_EPIC] Create/revise Factory class](https://trello.com/c/hJ3P6WHB/474-8-configepic-create-revise-factory-class)

virtwho/config.py
```diff
@@ -605,13 +573,6 @@ def _readConfig(self, parser):
...
-    def readFile(self, filename):
-        parser = StripQuotesConfigParser()
-        fname = parser.read(filename)
-        if len(fname) == 0:
-            self.logger.error("Unable to read configuration file %s", filename)
-        self._readConfig(parser)
-
     @staticmethod
     def map_destinations_to_sources(configs, dest_classes=(Satellite5DestinationInfo, Satellite6DestinationInfo)):
         """
```
jirihnidek, line 578: Doc string should be updated: `@rtype: (set, set, dict, set)`

virtwho/config.py
```diff
@@ -799,7 +776,7 @@ def iteritems(self):
             yield (key, self._values[key])
 
     def items(self):
-        return [(key, self[key]) for key in self]
+        return sorted([(key, self[key]) for key in self])
```
jirihnidek: Method `items` should be iterator to be compatible with Python3 and to bee little bit more memory effective.

virtwho/config.py
```diff
@@ -1096,51 +1104,54 @@ def filter_parameters(desired_parameters, values_to_filter):
...
             if conf.startswith('.'):
                 continue
-            try:
-                filename = parser.read(os.path.join(VW_CONF_DIR, conf))
-                if len(filename) == 0:
-                    logger.error("Unable to read configuration file %s", conf)
-            except MissingSectionHeaderError:
-                logger.error("Configuration file %s contains no section headers", conf)
-        return parser._sections
+            all_sections.update(parse_file(os.path.join(config_dir, conf), parser=parser))
```
jirihnidek: The default parses is ` StripQuotesConfigParser` so the `parses=parser` could be omitted here.

virtwho/config.py
```diff
@@ -1096,51 +1104,54 @@ def filter_parameters(desired_parameters, values_to_filter):
...
+    def all_drop_dir_config_sections(config_dir=VW_CONF_DIR):
         """
         Read all configuration sections in the default config directory
-        :return: a dictionary of {section_name: {key: value, ... } ... }
+        :return: a dictionary of {name: {key: value, ... } ... }
         """
         parser = StripQuotesConfigParser()
         all_dir_content = None
         conf_files = None
         non_conf_files = None
```
jirihnidek: Initialization of `all_dir_content', 'conf_files' and `non_conf_files` is useless.

virtwho/config.py
```diff
@@ -1075,10 +1069,24 @@ def validate(self):
             # This next check will not be necessary after we know that all sections are
             # ConfigSections
             if getattr(section, 'validate', None) is not None:
-                section.validate(validation_messages=validation_messages)
+                validation_messages.extend(section.validate())
         self.validation_messages = validation_messages
         return validation_messages
 
+    def validation_messages_as_string(self):
```
jirihnidek: It is not used at any place and any unit test doesn't exist for this method either.

virtwho/config.py
```diff
@@ -837,17 +814,25 @@ def __len__(self):
...
+        if 'virttype' in self._values:
+            self._values['type'] = self._values['virttype']
         # Finally calls _update_state
         self._update_state()
         return validation_messages
 
     def is_default(self, key):
         return self.defaults[key] == self._values[key]
 
+    def is_section_default(self):
```
jirihnidek, line 838: The name of method could be more self-descriptive or please add doc string.

virtwho/config.py
```diff
@@ -691,18 +651,35 @@ def parse_dests_from_dict(dict_to_parse,
...
-        logger.error("Unable to read configuration file %s", filename)
-    sections = parser._sections
+    parser = parser or StripQuotesConfigParser()
+    sections = {}
+    try:
+        fname = parser.read(filename)
+        if len(fname) == 0:
+            logger.error("Unable to read configuration file %s", filename)
+        else:
+            sections = parser._sections
```
jirihnidek: Is it really necessary to use `parses._sections`? I think it is possible to use `parses.sections()` and `parser.remove_section()` at line 708.

virtwho/config.py
```diff
@@ -1204,23 +1226,44 @@ def init_config(env_options, cli_options):
...
     global _effective_config
     _effective_config = effective_config
     effective_config.validate()
 
+    # Log pending errors
+    for err in validation_errors:
+        method = getattr(logger, err[0])
+        if method is not None:
+            method(err[1])
+
```
jirihnidek: This looks like familiar to me ;-).

virtwho/config.py
```diff
@@ -1204,23 +1226,44 @@ def init_config(env_options, cli_options):
...
+        validation_errors.append(('warning',
+                                  'Using default "%s" configuration' % VW_ENV_CLI_SECTION_NAME))
+        # In order to keep compatibility with older releases of virt-who,
+        # fallback to using libvirt as default virt backend
+        # only if we did not have a non_default env/cmdline config
+        if not has_non_default_env_cli:
+            effective_config[VW_ENV_CLI_SECTION_NAME] = ConfigSection.from_dict(
+                    DEFAULTS[VW_ENV_CLI_SECTION_NAME],
+                    VW_ENV_CLI_SECTION_NAME,
+                    effective_config)
```
jirihnidek: I hope this new code could be in some function.

virtwho/config.py
```diff
@@ -1236,3 +1279,8 @@ def get_effective_config():
     # We don't have one, let us make one
     _effective_config = EffectiveConfig()
     return _effective_config  # Initialize an EffectiveConfig with no options
+
+
+def reset_effective_config():
+    global _effective_config
+    _effective_config = None
```
jirihnidek: BTW: I would modify `EffectiveConfig` to be singleton. Using `global` is not my favorite approach. ;-)

jirihnidek (changes requested):
I think that it is on good way. I have some suggestions.

virtwho/config.py
cnsnyder: I'm not sure I understand? What is the request here?

virtwho/config.py
cnsnyder: ;-)

cnsnyder:
I have made the EffectiveConfig class a singleton through the use of a metaclass (/me cringes at the thought). Spent quite a bit of time playing with different options for accomplishing this and it seems this works the best for our use case. (Also updated to address the other comments).

Worth mentioning, in our case (thanks to pickle) we can still end up with different instances of the EffectiveConfig.

I look forward to hearing your thoughts.

Cheers!

virtwho/config.py
jirihnidek: I'm sorry. I meant the block of code between lines 1231-1252 could be refactored into some function (e.g. named `check_valid_conf()`). It is not good practice to have long functions (due to worse readability, etc.). The function `init_config()` is quite long now and it is not worthwhile to make it longer :-).

jirihnidek (changes requested):
I have only one comment containing answer to your question. Overall your other changes are good. 👍

jirihnidek (approved):
Ack :-)

PR #88, opened 2017-07-31
1211435: Don't send host-to-geust mapping, when env, owner are wrong

jirihnidek:
* Bug fix: https://bugzilla.redhat.com/show_bug.cgi?id=1211435
* When `env` or `owner` do not match values returned from
  satellite server, then the host-to-guest mapping is not
  reported and errror is logged.
* TODO: fix all remaining unit tests, when this change will
  be approved.


jirihnidek:
@cnsnyder I marked some tests to be skipped and many of complex tests are failing. Let me know, if host-to-guests mapping should not be reported, when `env` and `owner` in configuration are wrong. I will fix unit tests according your approval or disapproval/suggestions.

jirihnidek:
[Bug 1211435 - it should show error msg when env and owner is different from it on Satellite server](https://trello.com/c/1ANrORWJ/499-bug-1211435-it-should-show-error-msg-when-env-and-owner-is-different-from-it-on-satellite-server)

cnsnyder:
A few thoughts on this.
It seems upon further inspection, that the only truly valid configurations (when trying to point virt-who towards a Satellite or some other RHSM service) are one of the following:
1) Have the following configuration parameters:
    - rhsm_username  # for a user which is able to do hypervisor checkins for the owner below
    - rhsm_password
    - owner
    - env

OR

2) The system virt-who is running on is registered to the org we are trying to report to.

It is worth noting at this point that it seems virt-who will only attempt to use consumer cert-based auth is when one or both of the configuration parameters 'rhsm_username', 'rhsm_password' were not specified.

My understanding of the bug is that the reporter (for whatever reason) is not aware that the 'rhsm_username' and 'rhsm_password' args are required for virt-who to successfully use 'owner' and 'env'. The only 'owner' and 'env' that will work without the username and password will be the owner of the consumer which is represented by the consumer cert on the system.

I think the behaviour of virt-who (for now) should be as follows:
Only check the 'owner' and 'env' specified when using cert based auth (when 'rhsm_username' and or 'rhsm_password' have not been specified).
If the owner and env do not match, then do not send.

I'll point out specific changes in the files themselves as well.


Ideally, this sort of thing is something we'll check for in the future before execution gets to this point. But as a fix going into master I'd say something like the above is the way to go.

virtwho/manager/subscriptionmanager/subscriptionmanager.py
```diff
@@ -73,6 +75,33 @@ def readConfig(self):
...
+    def _check_owner_lib(self, config):
+        """
+        Try to check values of env and owner. These values has to be
+        equal to values obtained from Satellite server.
+        :param config: Configuration of virt-who
+        :return: None
+        """
+        if config is None:
+            return None
+
```
cnsnyder: I would continue with this method only if 'rhsm_username', 'rhsm_password' are not set and 'owner' and 'env' are set.

cnsnyder (changes requested):
See above requested changes and thoughts please!

jirihnidek:
@cnsnyder Thanks for review. I hope I did all changes you requested. Let me know, if it is OK. I will squash commits in the case.

cnsnyder (approved):

cnsnyder:
@jirihnidek LGTM! Will merge after squash!

jirihnidek:
@cnsnyder I squashed and pushed it to this branch.

PR #89, opened 2017-08-02
1408556: Log which owner updated mappings are being sent to

cnsnyder:
https://trello.com/c/nVHP23EF/513-3-bug-1408556-rfe-add-where-host-to-guests-mapping-is-pointing-to-in-the-rhsmlog

https://bugzilla.redhat.com/show_bug.cgi?id=1408556

### DIFF
--- a/virtwho/manager/subscriptionmanager/subscriptionmanager.py
+++ b/virtwho/manager/subscriptionmanager/subscriptionmanager.py
@@ -179,8 +179,9 @@ class SubscriptionManager(Manager):
 
         is_async = self._is_rhsm_server_async(report, connection)
         serialized_mapping = self._hypervisor_mapping(report, is_async, connection)
-        self.logger.debug("Host-to-guest mapping being sent to '%s': %s", report.config.owner,
-                          json.dumps(serialized_mapping, indent=4))
+        self.logger.debug("Host-to-guest mapping being sent to '{owner}': {mapping}".format(
+                          owner=report.config.owner,
+                          mapping=json.dumps(serialized_mapping, indent=4)))
 
         try:
             try:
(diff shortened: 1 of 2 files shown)


cnsnyder:
[[3] Bug 1408556 - [RFE] Add where host-to-guests mapping is pointing to in the rhsm.log](https://trello.com/c/nVHP23EF/513-3-bug-1408556-rfe-add-where-host-to-guests-mapping-is-pointing-to-in-the-rhsmlog)

virtwho/manager/subscriptionmanager/subscriptionmanager.py
```diff
@@ -179,7 +179,8 @@ def hypervisorCheckIn(self, report, options=None):
 
         is_async = self._is_rhsm_server_async(report, connection)
         serialized_mapping = self._hypervisor_mapping(report, is_async, connection)
-        self.logger.debug("Host-to-guest mapping: %s", json.dumps(serialized_mapping, indent=4))
+        self.logger.debug("Host-to-guest mapping being sent to '%s': %s", report.config.owner,
```
vritant: @cnsnyder not sure if this works, and I dont think all the counts are necessary. I think we can fix this in a single line change:

\-                self.logger.info('Hosts-to-guests mapping for config "%s": %d hypervisors and %d guests found',  report.config.name, hypervisor_count, guest_count)
\+                self.logger.info('Hosts-to-guests mapping for config "%s": %d hypervisors and %d guests found for %s',  report.config.name, hypervisor_count, guest_count, report.config.owner)


thoughts?

vritant (changes requested):

virtwho/manager/subscriptionmanager/subscriptionmanager.py
barnabycourt: You could also use str.format() with named parameters (https://docs.python.org/2/library/stdtypes.html#str.format,

cnsnyder: @vritant I believe we should still (at least on the debug level) be dumping the hypervisor-guest mapping being sent. In the location where we are doing the counting (and logging at the INFO level), I can make the change requested (and use str.format if so desired).

Because the reports are batched (and we have no name for the batched reports sent) we can't change just that line.

vritant:
@cnsnyder still doesnt work for me, lets chat online, perhaps I'm doing something wrong.

PR #92, opened 2017-08-09
Implemented ConfigSection subclass for virtual backends

jirihnidek:
* Implemented `VirtConfigSection` (subclass of ConfigSection) implementing most of validation logic from ecisting `Config` class. Logic specific for specific virtual backends were skipped. I assume this logic will be added into next subclasses.
* Implemented lot of unit test for new class
* Refactoring of `ConfigSection` and added bunch of unit tests for `ConfigSection`
* Method `validate()` is split to three methods: `_pre_validate()`,
  `_validate()` and `_post_validate()`. 'Post' and 'pre' methods
  are used for steps shared between subclasses. Method `_validate()`
  is intended for own implementation of validation in subclasses.
* Use attribute `invalidate_keys` and removed `to_remove`, because
  the `to_remove` duplicated functionality od `invalidate_keys`.
* Refactored method `_validate_configs()` as generic methods
  `_validate_list()` in `ConfigSection` and I'm reusing it in
  `VirtConfigSection` subclass
* Added support for `REQUIRED` options.


jirihnidek:
[[13] [CONFIG_EPIC] Create Base ConfigSection for virts](https://trello.com/c/qkGVKMFX/512-13-configepic-create-base-configsection-for-virts)

tests/test_config_section.py
```diff
@@ -0,0 +1,546 @@
...
+        # to trigger __setitem__ of virt_config
+        for key, value in LIBVIRT_SECTION_VALUES.items():
+            self.virt_config[key] = value
+
+    def test_validate_virt_type(self):
+        """
+        Test validation of supported types of virtualization backends
+        """
+        test_virt_types = list(VW_TYPES[:])
+        test_virt_types.extend('vmware,' 'kvm')
```
kahowell: Did you mean:
```python
test_virt_types.extend(['vmware', 'kvm'])
```
?

tests/test_config_section.py
```diff
@@ -0,0 +1,546 @@
...
+            else:
+                self.assertIsNone(result)
+                value = self.virt_config.get('type')
+                self.assertEqual(value, virt_type)
+
+    def test_validate_missing_virt_type(self):
+        """
+        Test validation of missing type of virtualization backend
+        """
+        del self.virt_config['type']
```
kahowell: I'd prefer if you just created `virt_config` instances for each test method (using a factory method), because this `del` statement modifies the instance member `self.virt_config`. It works because `setUp` resets `self.virt_config`, but this is not immediately obvious.

tests/test_config_section.py
```diff
@@ -0,0 +1,546 @@
...
+    def test_validate_unicode_unencrypted_password(self):
+        """
+        Test of validation of password that is not encrypted and it contains some
+        UTF-8 string.
+        """
+        self.virt_config['password'] = 'Příšerně žluťoučký kůň pěl úděsné ódy.'
+        result = self.virt_config._validate_unencrypted_password('password')
+        self.assertIsNone(result)
+
+    def mock_pwd_file(self):
```
kahowell: Don't like how this modifies `Password` globally. Would rather use `mock` module to accomplish this.

tests/test_config_section.py
```diff
@@ -0,0 +1,546 @@
...
+        result = self.virt_config._validate_filter('filter_hosts')
+        self.assertIsNone(result)
+
+
+class TestGlobalConfigSection(TestBase):
+    """
+    Test base for testing class GlobalSection
+    """
+
+    def setUp(self):
```
kahowell: Similar to above, I'd rather see this logic as a factory method.

tests/test_config_section.py
```diff
@@ -0,0 +1,546 @@
...
+        print(validate_messages)
+        expected_results = [
+            ('warning', 'Ignoring unknown configuration option "wrong_bool_value"'),
+            ('warning', 'log_per_config must be a valid boolean, using default. See man virt-who-config for more info'),
+            ('warning', 'Value for reporter_id not set, using default'),
+            ('warning', 'Value for log_dir not set, using default'),
+            ('warning', 'background must be a valid boolean, using default. See man virt-who-config for more info')
+        ]
+        self.assertGreater(len(validate_messages), 0)
+        # self.assertEqual(expected_results, validate_messages)
```
kahowell: Why is this commented out?

tests/test_config_section.py
```diff
@@ -0,0 +1,546 @@
...
+        """
+        Test the validate method will set default value, when an option
+        is missing and default value exist in DEFAULT
+        """
+        self.global_config.validate()
+        default_value = self.global_config.get('log_dir')
+        self.assertEqual(default_value, DEFAULT_LOG_DIR)
+
+
+MY_SECTION_VALUES = {
```
kahowell: The flow here is odd. Maybe put this test in a separate file?

tests/test_config_section.py
```diff
@@ -0,0 +1,546 @@
...
+                validation_messages.append(error)
+                self._invalid_keys.add(key)
+            self._unvalidated_keys.remove(key)
+
+        self.validation_messages.extend(validation_messages)
+
+
+class TestConfigSection(TestBase):
+    """
+    Class used for testing og ConfigSection
```
kahowell: typo: s/og/of

virtwho/config.py
```diff
@@ -759,6 +764,11 @@ class ConfigSection(collections.MutableMapping):
     __marker = object()
 
     def __init__(self, section_name, wrapper):
+        """
+        Initialization of ConfigSection instance
+        :param section_name: The name of section
+        :param wrapper: ?
```
kahowell: Let's fill in `:param wrapper`... @cnsnyder what do you suggest?

virtwho/config.py
```diff
@@ -819,18 +828,62 @@ def __len__(self):
...
+        self.reset_to_defaults()
         # Finally calls _update_state
         self._update_state()
-        return validation_messages
+
+    def _validate(self):
+        """
+        Own implementation should be done in subclasses
+        """
+        # raise RuntimeError("This method should never be called and should be implemented in subclass")
```
kahowell, line 874: Did you mean to uncomment this?

kahowell (changes requested):

kahowell:
Also, there are some newly failing tests (let's make sure we fix these before merging please):

As far as I can tell, only these are new to this branch (I compared against `config_refactor` branch):
```
tests/test_subscriptionmanager.py::TestSubscriptionManagerConfig::test_sm_config_cmd FAILED
tests/test_virtwho.py::TestOptions::test_default_cmdline_options FAILED
tests/test_virtwho.py::TestOptions::test_minimum_interval_options FAILED
tests/test_virtwho.py::TestOptions::test_options_debug FAILED
tests/test_virtwho.py::TestOptions::test_options_hierarchy_for_reporter_id FAILED
```

virtwho/config.py
cnsnyder: A reference to this ConfigSection's parent.

tests/test_config_section.py
jirihnidek: Yes, I did. Thanks.

tests/test_config_section.py
jirihnidek: You are right explicit is better than implicit in this case, especially for new developer trying to add another new test.

virtwho/config.py
jirihnidek: The instance of `ConfigSection` is currently used and uncommenting this would completely break virt-who ATM. I guess it is going to be changed soon and we will use only subclasses of `ConfigSection`. Then it will be possible uncomment this line, right? @cnsnyder

tests/test_config_section.py
jirihnidek: The format of message is still subject of changes. I will add TODO comment there.

virtwho/config.py
jirihnidek: @cnsnyder What exactly do you mean with `ConfigSection`'s parent? Is it parent class? In this case we can get it automatically using `self.__class__.__bases__`. Or do you mean other instance of ConfigSection (or it's subclass) referencing this ConfigSection (e.g. instance of LibVirtdConfigSection referencing to instance of VirtConfigSection?

tests/test_config_section.py
jirihnidek: @kahowell Do you mean move `MyConfigSection` to `stubs.py` and split `test_config_section.py` to three files: `test_config_section.py`, `test_global_config_section.py` and `test_virt_config_section.py`?

jirihnidek:
@kahowell there should not be newly failing tests now.

tests/test_config_section.py
cnsnyder: Sounds good to me!

tests/test_config_section.py
```diff
@@ -0,0 +1,587 @@
...
+        self.init_virt_config_section()
+        self.virt_config['password'] = 'Příšerně žluťoučký kůň pěl úděsné ódy.'
+        result = self.virt_config._validate_unencrypted_password('password')
+        self.assertIsNone(result)
+
+    def mock_pwd_file(self):
+        f, filename = tempfile.mkstemp()
+        self.addCleanup(os.unlink, filename)
+        Password.KEYFILE = filename
+        Password._can_write = MagicMock(retun_value=True)
```
cnsnyder: Probable typo "retun_value" -> "return_value".

virtwho/config.py
cnsnyder: @jirihnidek Your approach sounds reasonable to me.

cnsnyder (changes requested):
Overall looks good to me.
There are a few minor things I've mentioned above.
I think (before we're done with this work but not this PR) we should make sure not to fill the log file with so many log messages. But as you've said in comments that piece is something we'll refine as we go before merge to master.

The last thing is would ask (before merge) is to look and see about fixing the following tests (as the new fixes should allow them to pass with some changes methinks):

 - test_virt.py:TestVirtInclude

tests/test_config_section.py
jirihnidek: Yeah, but not my fault. Copy pasted from `test_password.py` :-)

virtwho/config.py
cnsnyder: The intention is for the wrapper to be a reference to the containing VWEffectiveConfig object.
If we do not feel the need for it, it could be removed.

virtwho/config.py
```diff
@@ -953,8 +962,8 @@ def get_defaults(cls):
 
     @classmethod
     def class_for_type(cls, virt_type):
-        clazz = ConfigSection
-        # TODO: Update to use actual subclasses as the are created
+        clazz = cls
+        # TODO: Update to use actual subclasses as they are created
```
jirihnidek: TODO: remove this TODO :-), because we will not to implement it .In theory `virt_type` would be needed during initialization of object and `__new__(cls, *args, **kwargs)` could return desired instance of subsclass.

cnsnyder: Agreed!

cnsnyder (approved):
Looks good to me.
Let's wait for Khowell on Monday.

tests/test_config_section.py
jirihnidek: I copy pasted this code from: `test_password.py` :-), but you are right. Modification of `Password` globally is not good idea. I added new method `unmock_pwd_file()` to restore `Password` to previous state. Look at: 1e84bc298c6276702673ea66caf49edbb5d72cf9 What do you think about this workaround?

kahowell: Much better. I think [`patch.multiple`](http://www.voidspace.org.uk/python/mock/patch.html#patch-multiple) would have worked as well, but this is fine.

kahowell (approved):

PR #95, opened 2017-08-28
1484265: Do not replace /etc/virt-who.conf on rpm upgrade

cnsnyder:
The /etc/virt-who.conf file was being overwritten on virt-who rpm upgrade.
This oneliner should fix that.

See: https://bugzilla.redhat.com/show_bug.cgi?id=1485865


mstead:
ACK

PR #99, opened 2017-10-04
Implement EsxConfigSection

kahowell:
Also refactored some ConfigSections to use a new method `add_key` which
makes the code a bit more DRY. Also the API is similar to argparse.

Changed the datastore to use deepcopy to avoid problems with pickling
config sections.


cnsnyder (changes requested):
I think that the validation for the server key for the esx config section should include the logic from line 125-126.

Other than that I think this is ready for merge

kahowell:
f9df395..0606733: Move logic for protocol to `_validate_server`

cnsnyder (approved):
LGTM! Thank you!

cnsnyder:
(Ignoring test failures as they are pre-existing) (Still on todo to fix)

PR #104, opened 2017-10-23
Create Xen Config Subclass

jirihnidek:
* Implemented subclass of VirtConfigSection for Xen virt. backend
* Implemented custom _validate_server() method for xen
* Fixed one bug in virt.py, when virt type unsupported is set in
  config file
* Fixed all unit tests in test_xen.py

### DIFF
--- a/virtwho/virt/xen/xen.py
+++ b/virtwho/virt/xen/xen.py
@@ -1,15 +1,69 @@
 #!/usr/bin/env python
 
-import sys
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+
 from time import time
 import XenAPI
 from XenAPI import NewMaster, Failure
 from collections import defaultdict
-import logging
 import requests
 
 from virtwho import virt
 from virtwho.util import RequestsXmlrpcTransport
+from virtwho.config import VirtConfigSection
+
+
+class XenConfigSection(VirtConfigSection):
+    """
+    This class is used for validation of Xen virtualization backend
+    section(s). It tries to validate options and combination of options that
+    are specific for this virtualization backend.
+    """
+    VIRT_TYPE = 'xen'
+    HYPERVISOR_ID = ('uuid', 'hostname')
+
+    def __init__(self, section_name, wrapper, *args, **kwargs):
+        super(XenConfigSection, self).__init__(section_name, wrapper, *args, **kwargs)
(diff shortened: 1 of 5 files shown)


jirihnidek:
[[8] [CONFIG_EPIC] Create Xen Config Subclass](https://trello.com/c/UfTXRH84/582-8-configepic-create-xen-config-subclass)

jirihnidek:
Please wait with review. I want to test this change with some Xen server.

jirihnidek:
I can confirm that this branch is able to get host-to-guest mapping from xen server.

virtwho/virt/xen/xen.py
```diff
@@ -1,15 +1,66 @@
...
 import requests
 
 from virtwho import virt
 from virtwho.util import RequestsXmlrpcTransport
+from virtwho.config import VirtConfigSection
+
+
+class XenConfigSection(VirtConfigSection):
+    """
+    This class is used for validation of fake virtualization backend
```
cnsnyder: 'fake' -> 'Xen'

virtwho/virt/xen/xen.py
```diff
@@ -1,15 +1,66 @@
...
+    section(s). It tries to validate options and combination of options that
+    are specific for this virtualization backend. In specific, it attempts to read
+    the given file and produces error messages if it is not usable.
+    """
+    VIRT_TYPE = 'xen'
+
+    def __init__(self, section_name, wrapper, *args, **kwargs):
+        super(XenConfigSection, self).__init__(section_name, wrapper, *args, **kwargs)
+
```
cnsnyder: Similar feedback as to the RHEVM PR #105 :

Please add something similar to the following as it seems 'server', 'username', and 'password' are required:

```
 self.add_key('server', validation_method=self._validate_server, required=True)
 self.add_key('username', validation_method=self._validate_username, required=True)
 self.add_key('password', validation_method=self._validate_unencrypted_password, required=True)
```

virtwho/virt/xen/xen.py
```diff
@@ -121,14 +168,14 @@ def getHostGuestMapping(self):
             if version:
                 facts[virt.Hypervisor.HYPERVISOR_VERSION_FACT] = version
 
-            if self.config.hypervisor_id == 'uuid':
+            if self.config['hypervisor_id'] == 'uuid':
```
cnsnyder, line 174: Should this be moved to a _validate_hypervisor_id method?

Would return an error message if the value provided was not one of those expected and valid for Xen.

cnsnyder (changes requested):
A few requested updates similar to those on the RHEVM PR.

Please take a look and let me know what you think!

Cheers,
Chris

virtwho/virt/xen/xen.py
jirihnidek: You are right. The `VirtConfigSection` has only lambda function returning `None` in all cases. I will fix it for other virt. backends (hyper-v and libvirtd) in this PR or I can create special PR when Xen and RHEVM fill be merged to config_refactor.

virtwho/virt/xen/xen.py
jirihnidek: True. I will fix this for hyper-v and libvirtd too and I will add some unit tests for this case.

virtwho/virt/xen/xen.py
jirihnidek: Solved in PR #105. When both PRs will be merged to config_refactor branch, then it should work as expected.

cnsnyder (approved):
Looks good to me!
Waiting to merge on #105
(Please squash before merge!)

PR #110, opened 2017-10-26
This fixes the remaining broken tests in the config_refactor branch!

cnsnyder:
What $subject said 😆

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 python:
   - "2.7"
-  - "2.6"
 env:
   - PYTHONPATH=.
 sudo: required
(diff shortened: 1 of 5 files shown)


cnsnyder:
I do not think we need to include a backport of the fix for python issue 1515 to get this to work on python 2.6 as well as 2.7. It appears that the behavior we rely on (that of properly copying instance methods), exists in python 2.6 on centos (and friends) but not on the version of 2.6 used by Travis...

To me it is either necessary to update the datastore (to not rely on deepcopy in the way that it is) OR we need to update the way that Travis is handling our 2.6 tests.

My preference is to update our testing infrastructure in the mid-term future.

Please do not merge until we have agreed on which route to proceed on (I don't want to get that "backport" in there (mainly used to help diagnose the issue)).

kahowell (approved):

PR #111, opened 2017-10-27
Jhnidek/remove old config

jirihnidek:
Completely removed `Config` from virt-who.


cnsnyder (approved):
LGTM! Thank you much!

PR #112, opened 2017-10-27
Merge the Config Refactor!!!!

cnsnyder:
There are many many changes here. I will focus on the high level pieces.
The Config class (for validating all configuration sections) has been replaced by a series of *ConfigSections.

These ConfigSections are mutable mappings that handle validation for a particular type of object in virt-who. More specifically, each source of host-guest mapping information now has validation specifically for it. They are generally named (Name_of_Virt_source)ConfigSection. They are generally located in each of the virt backend files.

There is a mutable mapping that contains all configuration sections that serves as the definitive source of configuration info inside virt-who. It is called EffectiveConfig. A method, init_config, layers and categorizes information from the sources we read (ENV, CLI, /etc/virt-who.conf, /etc/virt-who.d/*.conf) to build the effective config.

The effective config is then validated which produces a number of messages generated from each subsection. These messages come at various levels (mostly log levels like 'warning'). As of now, we only output those that are errors (will likely change this as a result of feedback from QE as we move forward).

Invalid configuration sections are removed.

Please Note:
I think it is worth considering squashing these commits down a bit. Please let me know in review if you agree.


tests/test_config_section.py
```diff
@@ -0,0 +1,399 @@
...
+    Class used for testing of ConfigSection
+    """
+
+    def __init__(self, *args, **kwargs):
+        super(TestConfigSection, self).__init__(*args, **kwargs)
+        self.my_config = None
+
+    def init_config_section(self):
+        """
+        This method is executed before each unit test
```
awood: Is this comment true?  If so, we could just put this functionality in `setup`.  If not, we should remove the comment.

tests/test_config_section_global.py
```diff
@@ -0,0 +1,182 @@
...
+    'interval': '120'
+}
+
+
+class TestGlobalConfigSection(TestBase):
+    """
+    Test base for testing class GlobalSection
+    """
+
+    def init_global_config_section(self):
```
awood: This looks like it is run every time before.  Does VirtWho not use `unittest.Testcase`?  The `setUp` method in `TestCase` seems like a good place to put this functionality so you don't have to call it by hand each time.

tests/test_config_section_global.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def test_validate_debug_value_bool(self):
+        """
+        Test validating of correct bool value 
+        """
+        self.init_global_config_section()
+        result = self.global_config._validate_str_to_bool('debug')
+        self.assertIsNone(result)
+        value = self.global_config.get('debug')
+        self.assertEqual(value, True)
```
awood: Just a thought: Depending on what you want, you may what to assert identity here instead of equality.  E.g.

```
In [1]: 1 == True
Out[1]: True

In [2]: 1 is True
Out[2]: False
```

virtwho/config.py
```diff
@@ -674,40 +337,1002 @@ def parse_dests_from_dict(dict_to_parse,
...
+        Steps necessary to do before evaluation 
+        """
+        if len(self._unvalidated_keys) == 0:
+            self.validation_messages.append(('warning', 'No values provided in: %s' % self.name))
+        else:
+            for default_key in self.defaults.keys():
+                if default_key not in self._unvalidated_keys:
+                    self.validation_messages.append(
+                        (
+                            'warning',
```
awood, line 554: Do we need to worry about internationalization at all?

awood (changes requested):
I would recommend squashing this down some by doing an interactive rebase and just squashing together consecutive commits that makes sense to be together.

virtwho/config.py
cnsnyder: No sir!

.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: python
 python:
   - "2.7"
-  - "2.6"
```
jirihnidek, line 4: We don't test it for 2.6 anymore? Why? We don't plan to use it at RHEL6?

tests/test_config_section.py
```diff
@@ -0,0 +1,397 @@
...
+        my_config.update(**MY_SECTION_VALUES)
+        for key, value in my_config.items():
+            self.assertEqual(value, MY_SECTION_VALUES[key])
+
+    # ---------- ADD_KEY AND RELATED VALIDATION TESTS ----------------------------------
+    # Note about the below tests:
+    # Some of the functionality of ConfigSection.validate (and add_key) is dependant on the
+    # values returned by the validation_method given. Also, at present, some of the
+    # validation_methods used in subclasses modify the values of the containing class (by
+    # accessing _values etc). In my opinion this should stop.
```
jirihnidek, line 182: What exactly is not wanted functionality? Do you mean the modification of values (e.g. modification of 'server' option widely used in most subclasses of `VirtConfigSection`). We already discussed it here: https://github.com/virt-who/virt-who/pull/97#discussion_r137029129 before first modification in `validate_*` method was implemented.

Your proposal can work, but it would make implementation little bit complicated and consider this. Validation of some option can be dependent on value of other valid (possibly modificated) option.

tests/test_config_section.py
```diff
@@ -0,0 +1,397 @@
...
+        self.init_config_section()
+
+    def init_config_section(self):
+        """
+        This method is executed before each unit test
+        """
+        self.my_config = MyConfigSection(MY_SECTION_NAME, None)
+        # Fill config with some values
+        for key, value in MY_SECTION_VALUES.items():
+            self.my_config[key] = value
```
jirihnidek, line 71: Because @kahowell asked for explicit initialization of `MyConfigSection`: https://github.com/virt-who/virt-who/pull/92#discussion_r133266049, I created this `init_config_section()` method, but it seems it was changed back to my idea of using `setUp()` :-). Move code from `init_config_section()` to `setUp()` or use explicit calling of `init_config_section()`.

tests/test_config_section_global.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def init_global_config_section(self):
+        """
+        Method executed before each unit test
+        """
+        self.global_config = GlobalSection('global', None)
+        # We need to set values using this way, because we need
+        # to trigger __setitem__ of global_config
+        for key, value in GLOBAL_SECTION_VALUES.items():
+            self.global_config[key] = value
```
jirihnidek, line 56: Same comment as for `init_config_section` in `test_config_section.py`.

tests/test_datastore.py
```diff
@@ -7,11 +7,11 @@
...
     def setUp(self):
-        pickle_patcher = patch('virtwho.datastore.pickle')
-        self.mock_pickle = pickle_patcher.start()
-        self.mock_pickle.dumps.return_value = sentinel.pickled_value
-        self.mock_pickle.loads.return_value = sentinel.unpickled_value
-        self.addCleanup(pickle_patcher.stop)
+        copy_patcher = patch('virtwho.datastore.copy')
+        self.mock_copy = copy_patcher.start()
+        self.mock_copy.deepcopy.side_effect = [sentinel.deep_copy_value_1,
+                                                sentinel.deep_copy_value_2]
```
jirihnidek, line 13: Wrong indentation.

BTW: confusing that you can use: `sentinel.any_strange_name_of_attribute` without raising exception.

tests/test_effective_config.py
```diff
@@ -0,0 +1,199 @@
...
+            'foo',
+            'bar',
+            'foo.bar'
+        ]
+        values = {
+            'foo': 'Foo',
+            'bar': 'Bar',
+            'bar.foo': 'Bar Foo',
+            'pub': 'Pub Foo'
+        }
```
jirihnidek, line 176: 'Foo', 'Bar', 'Pub' ... the developer implementing this should be more original next time ;-).

cnsnyder:
Closing in favor of the now merged #113

jirihnidek (commented):
I guess that some old code still passed in, because following command returns IMHO some obsolete code:

```
grep -R "config\." * 
```

PR #113, opened 2017-10-30
Merge config_refactor to master (v2)!!!

cnsnyder:
There are many many changes here. I will focus on the high level pieces.
The Config class (for validating all configuration sections) has been replaced by a series of *ConfigSections.

These ConfigSections are mutable mappings that handle validation for a particular type of object in virt-who. More specifically, each source of host-guest mapping information now has validation specifically for it. They are generally named (Name_of_Virt_source)ConfigSection. They are generally located in each of the virt backend files.

There is a mutable mapping that contains all configuration sections that serves as the definitive source of configuration info inside virt-who. It is called EffectiveConfig. A method, init_config, layers and categorizes information from the sources we read (ENV, CLI, /etc/virt-who.conf, /etc/virt-who.d/*.conf) to build the effective config.

The effective config is then validated which produces a number of messages generated from each subsection. These messages come at various levels (mostly log levels like 'warning'). As of now, we only output those that are errors (will likely change this as a result of feedback from QE as we move forward).

Invalid configuration sections are removed.

This is a rework of the history of the config_refactor branch destined for master. It should be compared to #112 to see if this history is better.


kahowell (approved):
ship it!

cnsnyder:
Merging based on above ack!

PR #116, opened 2017-11-13
1510760: Ensure virt-who exits properly (w/ no good conf)

cnsnyder:
Stops all use of sys.exit in the executor.
Now when the executor needs to exit, an ExitRequest is raised (with an appropriate exit code and message).

The exit function of virtwho.main ensures resources are destroyed as they need to be.
This function is called in response to an ExitRequest raised from the executor.

### DIFF
--- a/virtwho/executor.py
+++ b/virtwho/executor.py
@@ -1,6 +1,5 @@
 import time
 from threading import Event
-import sys
 
 from virtwho import log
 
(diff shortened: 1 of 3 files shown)


cnsnyder:
[Bug 1510760 - Virt-who process should exit when no suitable backend configuration can be found](https://trello.com/c/GapqSSWg/661-bug-1510760-virt-who-process-should-exit-when-no-suitable-backend-configuration-can-be-found)

jirihnidek (commented):
It seems good. I can find many `sys.exit()` in scripts of virt-who, but it is also used twice in `virtwho/virt/virt.py`. I suspect that it can be resource of similar problems. I vote for end running of virt-who at one point.

jirihnidek (approved):
ACK

PR #119, opened 2017-11-16
1511308: Xen and Hyper-V do not support some filter options

jirihnidek:
* Bug fix: https://bugzilla.redhat.com/show_bug.cgi?id=1511308
* Options: exclude_host_parent and filter_host_parents are not
  supported on Xen and Hyper-V mode
* Added more unit tests for XenConfigSection


jirihnidek:
[Bug 1511308 - No remind info in the log when set filter_host_parents or exclude_host_parents in xen/hyperv mode](https://trello.com/c/SJiWAWlb/665-bug-1511308-no-remind-info-in-the-log-when-set-filterhostparents-or-excludehostparents-in-xen-hyperv-mode)

cnsnyder (changes requested):
Might I suggest that we move the "filter_host_parents" and "exclude_host_parents" to the ESX config section and remove them from the VirtConfigSection class.

We can then let the base class include these options in the list of those that are unknown for the mode. (We might want to change the messaging there a little bit "%(keys)s unsupported in mode %(mode)s" or something similar).

cnsnyder (approved):
Looks good to me.
(Merging despite travis failures as those are not test failures, they are failures in travis setting up).

PR #120, opened 2017-11-16
1512778: ESX should require username, password, and server values

cnsnyder:
This modifies the EsxConfigSection to require the 'server', 'username', and 'password' options.

Esx requires these options to operate properly.

Bug fix: https://bugzilla.redhat.com/show_bug.cgi?id=1512778

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -71,8 +71,11 @@ class ConfigSectionValidationTests(object):
     # Should be changed for each set of tests, to make sure the right class is utilized
     CONFIG_CLASS = VirtConfigSection
 
-    def create_config(self):
-        return self.CONFIG_CLASS("test", None)
+    # should be a list of the keys which are required
+    REQUIRED_KEYS = set()
+
+    # Those keys that should have a default
+    KEYS_WITH_DEFAULTS = set()
 
     def _modified_dict(self, values, excluding=None, including=None):
         """
@@ -113,7 +116,8 @@ class ConfigSectionValidationTests(object):
         """
         config = self.CONFIG_CLASS.from_dict(self.VALID_CONFIG, "test", None)
 
-        self.assertIsInstance(config, self.CONFIG_CLASS, 'Wrong type of config, was "type" set correctly in "VALID_CONFIG"')
+        self.assertIsInstance(config, self.CONFIG_CLASS, 'Wrong type of config, was "type" set'
+                                                         'correctly in "VALID_CONFIG"?')
 
         # Having been updated, the config section needs validation
         self.assertEqual(config.state, ValidationState.NEEDS_VALIDATION)
@@ -133,7 +137,7 @@ class ConfigSectionValidationTests(object):
         """
         config = self.CONFIG_CLASS("test", None)
         # A sorted list of keys that are required, but have no default
-        target_keys = sorted(list(config._required_keys - set(config.defaults.keys())))
+        target_keys = sorted(list(self.REQUIRED_KEYS - self.KEYS_WITH_DEFAULTS))
         config_values = self._modified_dict(self.VALID_CONFIG, excluding=target_keys)
         config.update(**config_values)
 
@@ -151,7 +155,7 @@ class ConfigSectionValidationTests(object):
         NOTE: This does not account for modifications of default values made during validation.
         """
         config = self.CONFIG_CLASS("test", None)
-        target_keys = sorted(list(config._required_keys & set(config.defaults.keys())))
+        target_keys = sorted(list(self.REQUIRED_KEYS & self.KEYS_WITH_DEFAULTS))
         config_values = self._modified_dict(self.VALID_CONFIG, excluding=target_keys)
         config.update(**config_values)
 
--- a/tests/base.py
+++ b/tests/base.py
@@ -1,7 +1,7 @@
 """
 Basic module for tests,
 
-Copyright (C) 2014 Radek Novacek <rnovacek@redhat.com>
+Copyright (C) 2017 Radek Novacek <rnovacek@redhat.com>
 
 This program is free software; you can redistribute it and/or
 modify it under the terms of the GNU General Public License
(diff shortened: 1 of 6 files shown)


cnsnyder:
[Bug 1512778 - When configure ESX mode with username disabled and other valid configurations, virt-who will wait one interval time to send the valid one(s).](https://trello.com/c/gpBt2DbH/666-bug-1512778-when-configure-esx-mode-with-username-disabled-and-other-valid-configurations-virt-who-will-wait-one-interval-time-t)

jirihnidek (changes requested):
Please cover this case with some unit tests. There also other virt modes (Xen, Hyper-V, etc.) requiring server, username and password) and it is not cover by any unit test.

tests/base.py
```diff
@@ -51,3 +51,114 @@ def create_fake_config(name, **kwargs):
...
+            excluding = set(excluding)
+
+        if including is None:
+            including = set()
+        if isinstance(including, list):
+            including = set(including)
+
+        excluding = excluding - including
+
+        out = {}
```
jirihnidek, line 103: I would use dictionary comprehension here. Something like this:
```python
out = { key: value for key, value in values.items() if key in including or not including }
```

Then you can simplify following loop.

tests/test_config_section_esx.py
```diff
@@ -0,0 +1,41 @@
...
+    VALID_CONFIG = {
+        "type": "esx",
+        "server": "1.2.3.4",
+        "username": "username",
+        "password": "password",
+        "owner": "admin",
+        "env": "admin",
+        "filter_host_parents": "'PARENT_A', 'PARENT_B'",
+        "exclude_host_parents": "'PARENT_C_EXCLUDED'",
+    }
```
jirihnidek: No newline at end of file.

tests/base.py
```diff
@@ -51,3 +51,114 @@ def create_fake_config(name, **kwargs):
...
+
+    def test_valid_config(self):
+        """
+        This is essentially a smoke test to ensure the following:
+        1) An instance of the correct class is created from the valid configuration info
+        2) That the configuration is in the proper state pre and post validation
+        """
+        config = self.CONFIG_CLASS.from_dict(self.VALID_CONFIG, "test", None)
+
+        self.assertIsInstance(config, self.CONFIG_CLASS, 'Wrong type of config, was "type" set correctly in "VALID_CONFIG"')
```
jirihnidek: Please, wrap the line. There should be question mark at the end of message, right?

tests/base.py
```diff
@@ -51,3 +51,114 @@ def create_fake_config(name, **kwargs):
...
+    """
+
+    # A dictionary of valid configuration values
+    VALID_CONFIG = {}
+
+    # Should be changed for each set of tests, to make sure the right class is utilized
+    CONFIG_CLASS = VirtConfigSection
+
+    def create_config(self):
+        return self.CONFIG_CLASS("test", None)
```
jirihnidek: This method is never used, or did I miss anything?

tests/base.py
```diff
@@ -51,3 +51,114 @@ def create_fake_config(name, **kwargs):
...
+
+    def _modified_dict(self, values, excluding=None, including=None):
+        """
+        Copies a dict removing values where necessary
+        Args:
+            values: a dict of values to take from
+            excluding: a list or set of items to exclude from the copied dict
+            including: a list or set of items to include from the copied dict (wins over exclude)
```
jirihnidek, line 86: Argument `including` is never used, when this method is called. Is it really necessary to have it?

tests/base.py
```diff
@@ -51,3 +51,114 @@ def create_fake_config(name, **kwargs):
...
+        config = self.CONFIG_CLASS("test", None)
+        # A sorted list of keys that are required, but have no default
+        target_keys = sorted(list(config._required_keys - set(config.defaults.keys())))
+        config_values = self._modified_dict(self.VALID_CONFIG, excluding=target_keys)
+        config.update(**config_values)
+
+        messages = config.validate()
+
+        self.assertEqual(config.state, ValidationState.INVALID)
+        self.assertTrue(any(message[0] == 'error' for message in messages))
```
jirihnidek, line 147: This tests that there are some required keys missing. It would be nice to be able to set what keys are required. Something like `self.REQUIRED_KEYS`. Validity of unit tests are driven by code that you are testing. Required keys should be IMHO explicitly listed here in unit tests (in subclasses of `ConfigSectionValidationTests`).

tests/base.py
```diff
@@ -51,3 +51,114 @@ def create_fake_config(name, **kwargs):
...
+        config_values = self._modified_dict(self.VALID_CONFIG, excluding=target_keys)
+        config.update(**config_values)
+
+        config.validate()
+
+        self.assertEqual(config.state, ValidationState.VALID)
+
+        for key in target_keys:
+            self.assertIn(key, config, "Key '%s' is missing from config")
+            self.assertEqual(config[key], config.defaults[key])
```
jirihnidek: Similar comment here as for required keys. It would be nice to have some dictionary like this: `self.DEFAULT_VALUES`.

jirihnidek (changes requested):
I like the idea of `ConfigSectionValidationTests` 👍 , but I have some small requirements and suggestions :-).

tests/base.py
cnsnyder: Nope, left over from an earlier idea. Will remove.

tests/base.py
cnsnyder: Not necessarily, thought it might be nice to have in subclasses though.

tests/base.py
cnsnyder: Does python 2.6 support dict comprehensions?

jirihnidek: Yeah. Good point! It doesn't. Die, you ancient beast!. Die! :-)

tests/base.py
jirihnidek: OK, np :-)

tests/base.py
cnsnyder: I would agree but the defaults, as well as the required values are determined entirely by calls the add_key for the config class.

Do you think it would be enough to verify that the add_key method does what we expect? (and that the validation state of a config class given those calls to add_key does what we expect)?

jirihnidek: What about adding here something like this:

```python
self.assertEqual(config[key], self.DEFAULT_VALUES[key])
```

And subclass would contain something like this:

```python
DEFAULT_VALUES = { 'is_hypervisor': True }
```

jirihnidek (approved):
ACK

PR #121, opened 2017-11-17
1510310: Ensure that owner and env are required where necessary

cnsnyder:
Including env_vars.

Bug fix: https://bugzilla.redhat.com/show_bug.cgi?id=1510310


cnsnyder:
[Bug 1510310 - When disable the owner and env options in /etc/sysconfig/virt-who, host-to-guest info still can be sent to Satellite6.2](https://trello.com/c/Ozr6G90Y/667-bug-1510310-when-disable-the-owner-and-env-options-in-etc-sysconfig-virt-who-host-to-guest-info-still-can-be-sent-to-satellite62)

virtwho/__init__.py
```diff
@@ -3,3 +3,6 @@
 DefaultInterval = 3600  # One per hour
 MinimumSendInterval = 60  # One minute
 MinimumJobPollInterval = 15
+
+SAT5 = "satellite"
+SAT6 = "sam"
```
jirihnidek: Missing empty line at end of file.

virtwho/config.py
```diff
@@ -825,8 +859,7 @@ class VirtConfigSection(ConfigSection):
     def __init__(self, section_name, wrapper):
         super(VirtConfigSection, self).__init__(section_name, wrapper)
         self.add_key('type', validation_method=self._validate_virt_type, default='libvirt')
-        self.add_key('virttype', validation_method=self._validate_virt_type, default='libvirt',
-                     destination='virttype')
+        self.add_key('sm_type', validation_method=self._validate_sm_type, default="sam", restricted=True)
```
jirihnidek, line 862: The meaning of `restricted` argument is not clear to me. How is it restricted?

BTW: It is not covered by any unit test.

'sm_type' is marked as restricted at line 490 using different way.

virtwho/config.py
```diff
@@ -873,9 +906,30 @@ def __setitem__(self, key, value):
     def _pre_validate(self):
         # Used to allow virttype to be provided for type (if type was not given)
         if 'virttype' in self._values and 'type' not in self._values:
-            self._values['type'] = self._values['virttype']
+            self['type'] = self._values['virttype']
```
jirihnidek: It seem you try to do some refactoring 👍 (renaming variables), but I'm I'm little bit confused. Should it be `virttype` or `virtType` or `virt_type`? I can still find three names of key/variable for probably the same thing. Please finish your refactoring effort :-).

virtwho/config.py
```diff
@@ -1094,7 +1148,7 @@ def _validate(self):
 DEFAULTS = {
     VW_GLOBAL: GlobalSection.get_defaults(),
     VW_ENV_CLI_SECTION_NAME: {
-        'smtype': 'sam',
+        'sm_type': 'sam',
```
jirihnidek: Yes, only `sm_type` 👍

jirihnidek (changes requested):
I tested your PR and it seems that it fixes the problem. 👍 , but I request some changes. :-)

The concept of restricted keys is not clear to me.

I aware that the type of destination influences validation of configuration (especially owner, env), but IMHO it should be done inside validation method of e.g. `EsxConfigSection` not in `VirtConfigSection`. Look at condition at line 1069 in file config.py.

BTW: I don't like concept of using sysconfig and environment variables for options of virt. backends. Are we going to mark it as obsolete too?

cnsnyder:
I would like to move away from using environment variables for configuration as well. I think we'll need further discussions in order to make that happen though.

virtwho/config.py
cnsnyder: Restricted keys are meant to be those which need to be validated (and otherwise included in the scheme of things) but are considered internal meaning we should not output validation messages about them.

virtwho/config.py
cnsnyder: Will do!

jirihnidek:
@cnsnyder, you added unit tests. Great! 👍 What about one name for: `virttype` or `virtType` or `virt_type`?

jirihnidek (approved):
ACK

PR #123, opened 2017-12-15
1522383: Remove global option background

jirihnidek:
* Bugfix: https://bugzilla.redhat.com/show_bug.cgi?id=1522383
* Removed background option, because it is not supported anymore.
* Fixed unit tests


jirihnidek:
[Bug 1522383 - No global option "background" exists, but it always prints warning info about the option not set.](https://trello.com/c/mXr9wmDn/705-bug-1522383-no-global-option-background-exists-but-it-always-prints-warning-info-about-the-option-not-set)

cnsnyder (changes requested):
We still need to be able to run as a daemon for older versions of Centos (6) and similar.
The fix we need for this bug is to not output a warning if the background option is not defined (wherever it might come from).

jirihnidek:
@cnsnyder It is true that `export VIRTWHO_BACKGROUND=1` is used in init script. I missed this, but `daemon` bash function used for daemonize the virt-who proces. I think we can still remove the background option. I will do experiments on RHEL 6 and do further modifications as needed.

jirihnidek:
@cnsnyder I can confirm that virt-who works without background option as expected on RHEL 6. I will remove remaining notes about background (init script, tests, etc.) and then I will do rebase of this branch. OK?

cnsnyder:
@jirihnidek Sounds good to me! Thank you!

cnsnyder (approved):

PR #125, opened 2018-01-19
1509597: Enable to use virt-who wih VDSM again

jirihnidek:
* Bug fix: https://bugzilla.redhat.com/show_bug.cgi?id=1509597
* It wasn't able to use virt-who with VDSM due to refactoring
  of virt-who config
* Refactored vdsm.py a little
* Modified coresponding unit test a little

PR #124, opened 2017-12-21
Allow virt-who to run on either python 3.X or 2.X

cnsnyder:
* Includes complete compatibility of virt-who with python 3
* There is no expected change in virt-who behavior from prior
* Includes spec file updates to allow virt-who to install using the appropriate version of python (and to the appropriate locations.

Will need a squash, post-review.

PR #126, opened 2018-01-19
Add JSON-RPC support for VDSM

kahowell:


kahowell:
FYI, this has some overlap with #125 and probably some conflicts with #124. I'll happily rebase against either.

kahowell:
I added a main function to one of the new modules for smoketest of VDSM over JSON-RPC that works like this:

    python -m virtwho.virt.vdsm.jsonrpc --cert /etc/pki/vdsm/certs/vdsmcert.pem --key /etc/pki/vdsm/keys/vdsmkey.pem $HOSTNAME 54321 Host.getVMList onlyUUID=False

virtwho/virt/vdsm/stomp.py
```diff
@@ -0,0 +1,165 @@
...
+                message = StompFrame('CONNECT', headers={
+                    'accept-version': '1.2',
+                    'host': self.host,
+                })
+                self.socket.sendall(message.to_bytes())
+                self._recv_frame()  # consume CONNECTED message TODO verify message
+                return  # success connecting
+            except socket.error as e:
+                log.warning('Unable to connect %s:%s: %s' % (self.host, self.port, unicode(e)))
+        raise ConnectionError('Unable to connect to %s:%s' % (self.host, self.port))
```
jirihnidek: The `ConnectionError` is not defined nor imported.

virtwho/virt/vdsm/stomp.py
```diff
@@ -0,0 +1,165 @@
...
+    def _recv_frame(self):
+        while b'\x00' not in self.buffer:
+            self.buffer += self.socket.recv(4096)
+        frame, self.buffer = self.buffer.split(b'\x00', 1)
+        # add null byte back to frame
+        frame += b'\x00'
+        return frame
+
+    def subscribe(self, destination):
+        id = uuid.uuid4()
```
jirihnidek: Variable `id` shadows built-in method.

virtwho/virt/vdsm/stomp.py
```diff
@@ -0,0 +1,165 @@
...
+                })
+                self.socket.sendall(message.to_bytes())
+                self._recv_frame()  # consume CONNECTED message TODO verify message
+                return  # success connecting
+            except socket.error as e:
+                log.warning('Unable to connect %s:%s: %s' % (self.host, self.port, unicode(e)))
+        raise ConnectionError('Unable to connect to %s:%s' % (self.host, self.port))
+
+    def close(self):
+        for id in self.subscription_ids:
```
jirihnidek: Variable `id` shadows built-in method.

virtwho/virt/vdsm/vdsm.py
```diff
@@ -119,29 +122,61 @@ def _secureConnect(self):
...
+            except Exception:
+                self.jsonrpc_client = None
+                self.logger.exception('Unable to connect to JSON-RPC with SSL')
+                self.xmlrpc_client = self._secureConnect(addr, ssl_context)
                 return
-            except Exception as e:
-                self.logger.exception(e)
-        # Try http version if ssl is off or fails
-        self.server = xmlrpclib.Server("http://localhost:%s" % self.management_port)
+            except Exception:
```
jirihnidek: Broad exceptions are not good and the `Exception` has already been caught at line 139.

virtwho/virt/vdsm/stomp.py
```diff
@@ -0,0 +1,165 @@
...
+    def connect(self):
+        for family, _, _, _, addr in socket.getaddrinfo(self.host, self.port):
+            try:
+                self.socket = socket.socket(family)
+                if self.ssl_context:
+                    if hasattr(self.ssl_context, 'wrap_socket'):
+                        self.socket = self.ssl_context.wrap_socket(self.socket)
+                    else:
+                        self.socket = M2Crypto.SSL.Connection(self.ssl_context)
+                self.socket.connect(addr)
```
jirihnidek: You use socket in blocking mode. When vdsm server is not running/available, then smoke test will be blocked forever and behavior of virt-who will be strange.

I would recommend to use module `requests` for communication with server. Module `requests` is already used in `rhevm.py`. Thus using of `requests` here would not create another rpm dependency. Using of bare `socket` has potential for many problems.

virtwho/virt/vdsm/stomp.py
```diff
@@ -0,0 +1,165 @@
...
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""
+
+import logging
+import socket
+import uuid
+import M2Crypto.SSL.Connection  # TODO make conditional for Python 3 compat
+
+log = logging.getLogger(__name__)
+
```
jirihnidek: Please add here comment about `python-stomper` module. May be it will be possible to use it in future.

virtwho/virt/vdsm/stomp.py
```diff
@@ -0,0 +1,165 @@
...
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""
+
```
jirihnidek: Please add some doc string here with description of the module (what is STOMP, why is it used, ...).

virtwho/virt/vdsm/stomp.py
```diff
@@ -0,0 +1,165 @@
+from __future__ import unicode_literals
+
+"""
+Module for STOMP messaging, part of virt-who
+
+Copyright (C) 2018 Kevin Howell <khowell@redhat.com>
+
```
jirihnidek: GPL should not be doc string but it should be comment.

virtwho/virt/vdsm/jsonrpc.py
```diff
@@ -0,0 +1,115 @@
+from __future__ import unicode_literals
+
+"""
+Module for JSON-RPC over STOMP, part of virt-who
+
+Copyright (C) 2018 Kevin Howell <khowell@redhat.com>
```
jirihnidek: GPL license should not be part of doc string, but it should be in comment.

virtwho/virt/vdsm/jsonrpc.py
```diff
@@ -0,0 +1,115 @@
...
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""
+
+import json
+import ssl
+import uuid
+
+from virtwho.virt.vdsm.stomp import StompClient
+
```
jirihnidek, line 34: I suspect you aware of `python-jsonrpclib` module and I understand reasons why it is not used. Please add note (link) here about the module with explanation why it is not used.

jirihnidek (changes requested):
I can confirm that it just works :-). You are fast 👍 
On the other side I'm not happy that we cannot use existing modules e.g. `python-jsonrpclib`, but we probably cannot do anything about this. Please add some unit tests covering new code in `jsonrpc.py` and `stomp.py`. I also request some changes in code and I have some suggestions.

virtwho/virt/vdsm/stomp.py
kahowell: Unfortunately, STOMP is its own protocol, and does not use HTTP, so we can't use requests for this. I will investigate using a timeout.

virtwho/virt/vdsm/stomp.py
kahowell: Ah, it's a builtin in python3... whoops. I will switch to `IOError`.

kahowell:
@jirihnidek I have addressed all feedback and rebased against master.

There are several ways to view the commits in the PR...

Both of these, you've seen the changes, but are good for sanity-checking the rebase:
 - `git diff 5fbb71b..568f131` shows the difference between the PR's original commits before rebase and after rebase
 - `git diff 87f9cfe..568f131` shows the changes the original commits rebased makes to master.

Following that, I made a few logical commits addressing difference pieces:
 - `git show 76c11c9` documentation changes
 - `git show 3b110c3` refactoring to address XML-RPC fallback and fixing XML-RPC to work on modern Python (F27)
 - `git show 11f0062` addition of timeout
 - `git show 9e8cf5d` addition of tests for JSON-RPC and STOMP code, plus a few fixes

virtwho/virt/vdsm/stomp.py
```diff
@@ -0,0 +1,242 @@
...
+        Sends a STOMP CONNECT and consumes the subsequent CONNECTED message.
+
+        :raises: IOError if no connection attempts succeed.
+        """
+        for family, _, _, _, addr in socket.getaddrinfo(self.host, self.port):
+            try:
+                self.socket = socket.socket(family)
+                self.socket.settimeout(self.timeout)
+                if self.ssl_context:
+                    self.socket = self.ssl_context.wrap_socket(self.socket)
```
jirihnidek: This raises error on RHEL 7, because `self.ssl_context.wrap_socket` is not available here :-(

virtwho/virt/vdsm/vdsm.py
```diff
@@ -107,46 +112,120 @@ def _get_local_vds_name(self, trusted_store_path):
...
+            transport = None
+            if ssl_context:
+                if self._need_m2crypto():
+                    transport = xmlrpclib.SafeTransport(context=ssl_context)
+                else:
+                    from M2Crypto.m2xmlrpclib import SSL_Transport
+                    transport = SSL_Transport(ssl_context)
+                uri = 'https://%s:%s' % (addr, self.management_port)
+            else:
+                uri = 'http://%s:%s' % (addr, self.management_port)
```
jirihnidek, line 191: It looks like that we are enabling unsecured connection to remote server here. On the other side it seems that `_xmlrpc()` is called only with `addr='localhost'`, when `ssl_context=None`.

virtwho/virt/vdsm/vdsm.py
```diff
@@ -107,46 +112,120 @@ def _get_local_vds_name(self, trusted_store_path):
...
+            ctx.load_verify_locations(ca_cert)
+            ctx.load_cert(cert_file, key_file)
+        return ctx
 
     def connect(self):
         if self.trust_store_path:
-            try:
-                self.server = self._secure_connect()
+            ssl_context = self._create_ssl_context()
+            addr = self._get_local_vds_name(self.trust_store_path)
```
jirihnidek, line 158: It seems that `addr` is get only from certificate and value of `server` option from configuration file is completely ignored. This logic was present in old xmlrpc code too. @cnsnyder, I don' think that this is expected behavior. Thoughts?

jirihnidek (changes requested):
I like your progress 👍 , but unfortunately it doesn't work on RHEL7.  Details are in comments.

virtwho/virt/vdsm/stomp.py
kahowell: Had logic in for this, and removed unintentionally. I added it back in latest commit.

jirihnidek (approved):
ACK, please squash and then it is ready for merge.

cnsnyder:
One request for the squash, please add the bug number (1353119) to the beginning of the commit message. Thank you!

PR #127, opened 2018-01-26
1520236: Do not log traceback, when server returns 429 http error

jirihnidek:
* Bug fix: https://bugzilla.redhat.com/show_bug.cgi?id=1520236
* When rate limit exceeded, then catch this exception and reraise
  different exception. This is handled properly in virt.py
* Raised log level for this case from debug to info, because
  user should be informed why is communication broken with
  candlepin server.

### DIFF
--- a/virtwho/virt/virt.py
+++ b/virtwho/virt/virt.py
@@ -708,15 +708,18 @@ class DestinationThread(IntervalThread):
                         retry = False
                     except ManagerThrottleError as e:
                         if self._oneshot:
-                            self.logger.debug('429 encountered when sending virt guests in '
-                                              'oneshot mode, not retrying')
+                            self.logger.info(
+                                'Rate limit exceeded while sending host-guest mapping, please see: '
+                                'https://access.redhat.com/solutions/2212941. '
+                                'Running in one-shot mode, not retrying.')
                             sources_erred.append(source_key)
                             break
                         num_429_received += 1
                         retry_after = self.handle_429(e.retry_after, num_429_received)
-                        self.logger.debug('429 encountered when sending virt '
-                                          'guests.'
-                                          'Retrying after: %s', retry_after)
+                        self.logger.info(
+                            'Rate limit exceeded while sending host-guest mapping, please see: '
+                            'https://access.redhat.com/solutions/2212941. '
+                            'Retrying after: %s seconds.', retry_after)
                         self.wait(wait_time=retry_after)
                     except (ManagerError, ManagerFatalError):
                         self.logger.exception("Fatal error during send virt "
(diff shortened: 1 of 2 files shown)


jirihnidek:
This bug fix can be tested using `ncat`. It was described by @awood int his PR: https://github.com/candlepin/subscription-manager/pull/1759 . All you need to change in this case is create `429.response` file:

```
HTTP/1.1 429 Too Many Requests
Retry-After: 3600
Server: Apache-Coyote/1.1
x-candlepin-request-uuid: 9e62bc70-811f-4047-bf16-b7aeff20f593
X-Version: 2.0.43-1
X-Version: 2.0.43-1
Content-Type: application/json
Content-Length: 3

{}
```
and modify usage of `ncat` command a little bit:

```
ncat --listen --keep-open -vv --ssl-cert ./poor-man.crt --ssl-key ./poor-man.key -p 3131 --exec '/bin/cat 429.response'
```

jirihnidek:
[Bug 1520236 - When receive 429 info for stage candlepin, virt-who show RateLimitExceededException](https://trello.com/c/lzmBMgXU/741-bug-1520236-when-receive-429-info-for-stage-candlepin-virt-who-show-ratelimitexceededexception)

virtwho/virt/virt.py
```diff
@@ -708,15 +708,15 @@ def _send_data(self, data_to_send):
                         retry = False
                     except ManagerThrottleError as e:
                         if self._oneshot:
-                            self.logger.debug('429 encountered when sending virt guests in '
-                                              'oneshot mode, not retrying')
+                            self.logger.info('429 encountered when sending virt guests in '
```
cnsnyder: Is there a knowledge base article we can link to here?
The messaging is still a bit unclear if intended for an end user and not a dev.
I would suggest something like "Rate limit exceeded while sending host-guest mapping, please see {link to knowledge base article}" (optionally with the following appended in oneshot mode) " running in oneshot mode, not retrying".

Similar for the below log message as well.

cnsnyder (changes requested):
Works well for me as far as traceback removal.
I believe we should try to find a knowledge base article to link to if there is one (if not to create one then link). This would be the most informative for an end user seeing this sort of error.

virtwho/virt/virt.py
jirihnidek: I found this article: https://access.redhat.com/solutions/2212941, but it would be IMHO necessary to extend examples of `rhsm.log` error messages in this article or make modified copy of this article.

jirihnidek:
Created BZ bug report: https://bugzilla.redhat.com/show_bug.cgi?id=1540204 for subscription-manager related to this PR.

jirihnidek:
Created PR: https://github.com/candlepin/subscription-manager/pull/1771 for `rhsm` module related to this PR.

jirihnidek:
@cnsnyder please can you review this PR once again? The PR for subman: https://github.com/candlepin/subscription-manager/pull/1771 (blocking this PR) is already merged to master. Thanks.

cnsnyder (approved):
Looks great! Thank you much!

cnsnyder:
Please squash, then I'll merge!

jirihnidek:
Squashed and ready for merge.

PR #128, opened 2018-01-31
1492074: Enable login to ESX using password with UTF-8 string

jirihnidek:
* Bug fix: https://bugzilla.redhat.com/show_bug.cgi?id=1492074
* Module suds does not allow to use password as latin string or
  unicode. It does not allow to use string with UTF-8 characters.

### DIFF
--- a/virtwho/virt/esx/esx.py
+++ b/virtwho/virt/esx/esx.py
@@ -339,7 +339,11 @@ class Esx(virt.Virt):
         try:
             # Don't log message containing password
             logging.getLogger('suds.client').setLevel(logging.CRITICAL)
-            self.client.service.Login(_this=self.sc.sessionManager, userName=self.username, password=self.password)
+            self.client.service.Login(
+                _this=self.sc.sessionManager,
+                userName=unicode(self.username, 'utf-8'),
+                password=unicode(self.password, 'utf-8')
+            )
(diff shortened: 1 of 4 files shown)


jirihnidek:
[Bug 1492074 - Problem with virt-who using Unicode characters in a hypervisor's account password](https://trello.com/c/dBc83OMs/749-bug-1492074-problem-with-virt-who-using-unicode-characters-in-a-hypervisors-account-password)

virtwho/virt/esx/esx.py
```diff
@@ -339,7 +339,11 @@ def login(self):
         try:
             # Don't log message containing password
             logging.getLogger('suds.client').setLevel(logging.CRITICAL)
-            self.client.service.Login(_this=self.sc.sessionManager, userName=self.username, password=self.password)
+            self.client.service.Login(
+                _this=self.sc.sessionManager,
+                userName=self.username,
```
Ceiu: Do we need/want to do the same to both username and password here?

jirihnidek: Good point 👍 . It is IMHO not allowed to use username including any non-ascii character by ESX and virt-who should log some warning, when wrong username is used, but there was bug in checking of configuration file. I will update this PR.

cnsnyder (approved):
Looks good to me. Thank you!

PR #130, opened 2018-02-13
1447022: Log warning, when wrong filter is in config file

jirihnidek:
* Bug fix: https://bugzilla.redhat.com/show_bug.cgi?id=1447022
* When hypervisor_id is hostname or hwuuid, then filter should
  not contain any UUID, because no UUIDs are gathered in this
  case and filters will not work as expected.

### DIFF
--- a/virtwho/config.py
+++ b/virtwho/config.py
@@ -23,6 +23,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 import collections
 import six
 import os
+import uuid
 
 from six.moves.configparser import SafeConfigParser, NoOptionError, Error, MissingSectionHeaderError
 from virtwho import log, SAT5, SAT6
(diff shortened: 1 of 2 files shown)


jirihnidek:
[[5] Bug 1447022 - virt-who doesn't collect the information when 'hypervisor_id=hostname' is set and hosts are filtered by UUID](https://trello.com/c/a6xmvt6E/769-5-bug-1447022-virt-who-doesnt-collect-the-information-when-hypervisoridhostname-is-set-and-hosts-are-filtered-by-uuid)

jirihnidek:
@cnsnyder I'm considering now removing UUIDs from filter list, when `hostname_id != uuid`. Thoughts?

virtwho/config.py
```diff
@@ -1093,9 +1094,40 @@ def _validate_owner(self, key):
 
     def _validate_filter(self, filter_key):
         """
-        Try to validate filter option
+        Try to validate filter option. It can contain list of hostname and UUIDs
+        :param list_key: key of filter ('filter_hosts', 'exclude_hosts')
```
kahowell: Nitpick: param name is filter_key...

virtwho/config.py
```diff
@@ -1093,9 +1094,40 @@ def _validate_owner(self, key):
...
+                try:
+                    uuid.UUID(filter_value)
+                except ValueError:
+                    pass
+                else:
+                    wrong_filter_values.append(filter_value)
+            if len(wrong_filter_values) > 0:
+                result = (
+                    'warning',
+                    'Filter values: "%s" are not compatible with hypervisor_id = "%s"' %
```
kahowell: I think this warning could be more informative... Perhaps something like

```
Filter values: "%s" appear to be UUIDs. UUIDs are not gathered when hypervisor_id = "%s"
```

@cnsnyder thoughts?

kahowell (changes requested):
> @cnsnyder I'm considering now removing UUIDs from filter list, when hostname_id != uuid. Thoughts?

It is entirely possible (though somewhat odd) for the user to use UUIDs as hostnames; for example, the user executes `hostnamectl set-hostname $(uuidgen)`, and it would be surprising to such a user that they cannot filter hosts because of their naming convention...

virtwho/config.py
cnsnyder: I agree with this suggestion.

virtwho/config.py
jirihnidek: Good point.

virtwho/config.py
jirihnidek: Sounds good.

kahowell (approved):

kahowell:
@jirihnidek ready for squash. There was a failed test during PR travis runs but looks completely unrelated to changes from this PR, and works for me locally, so I'm hoping squashed run will be fine.

PR #135, opened 2018-03-26
1560461: virt-who reporting RHEVM hypervisor requires owner option

jirihnidek:
* Bug fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1560461

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -72,8 +72,11 @@ class ConfigSectionValidationTests(object):
     # Should be changed for each set of tests, to make sure the right class is utilized
     CONFIG_CLASS = VirtConfigSection
 
-    # should be a list of the keys which are required
-    REQUIRED_KEYS = set()
+    # should be a list of the keys which are required for sam destination
+    SAM_REQUIRED_KEYS = set()
+
+    # should be a list of the keys which are required for sam destination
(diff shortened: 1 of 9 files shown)


jirihnidek:
May be this problem is not RHEVM specific and we can expect similar problems with other hypervisors too, because REST API: http://www.candlepinproject.org/swagger/?url=candlepin/swagger-2.2.0.json#!/hypervisors/hypervisorUpdateAsync Thoughts?

cnsnyder:
@jirihnidek The problem, ultimately, is we need to account for the different destination types explicitly as per the original design.
If we are not sending information to Sat 6 or RHSM, (so if it's Sat 5), we do not need to know what the owner and env are.

I might suggest adding a kwarg to the add_key method, to specify if a key is required for Sat6 or Sat5. (or both).

cnsnyder (changes requested):
I would request that either a way of differentiating which destination type the owner and env params are required for be added as a modification of add_key, or that the validation of owner and env keys be updated to check for the type of destination being reported to.

(See above comment for more detail)

cnsnyder (approved):
LGTM, thank you for the updates!

PR #136, opened 2018-04-03
1554228: Unicode issue on status update call

wottop:
How to test this issue:

Start on master branch and use the  QE instance of rhevm4. Ask for the credentials if you need them.

Run the following as root [not sudo]
run 'virt-who -do', it should work properly.
run 'export RHSM_USE_M2CRYPTO=true'
run 'virt-who -do', it should fail as is described in the bug.

Change to this branch
run 'virt-who -do', it should work properly.


cnsnyder (changes requested):
One small request, could you update the commit message with just a few details of why this is necessary?

Will merge after that!

cnsnyder (approved):

PR #137, opened 2018-04-24
Update tito releasers to include newer versions of fedora

cnsnyder:
This updates the branches into which we'd like to release virt-who for fedora.
We should be targeting the same as subscription-manager in my opinion.


wottop:
ack

PR #138, opened 2018-04-24
1569299: try/exception needed for hypervisor_id check

wottop:
Otherwise the process of reading system data fails if only one
 system does not have a value for the field designated as the
 hypervisor id.


cnsnyder (approved):
LGTM! Thank you!

PR #139, opened 2018-04-24
1560598: Pass hostname to M2Crypto in STOMP client

kahowell:


virtwho/virt/vdsm/stomp.py
```diff
@@ -173,10 +173,11 @@ def connect(self):
                 if self.ssl_context:
                     if hasattr(self.ssl_context, 'wrap_socket'):
                         self.socket = self.ssl_context.wrap_socket(self.socket)
+                        self.socket.connect(addr)
                     else:
                         import M2Crypto.SSL.Connection
                         self.socket = M2Crypto.SSL.Connection(self.ssl_context)
-                self.socket.connect(addr)
+                        self.socket.connect((self.host, int(self.port)))
```
wottop, line 180: excess parenthesis?

wottop (changes requested):
How do I test this?

virtwho/virt/vdsm/stomp.py
kahowell: Actually, no. `connect` takes a tuple: see https://docs.python.org/2/library/socket.html

kahowell:
> How do I test this?

@wottop, you need:
 * a new enough version of RHV to support JSON-RPC over STOMP.
 * ensure that the RHV server has a certificate which uses the hostname (and *not* IP).
 * a version of python that does not support `wrap_socket` (see https://docs.python.org/2/library/ssl.html#ssl.SSLContext.wrap_socket). Looks like a python version older than 2.7.9 *should* work.

It might be easier to fake the python version requirement by tweaking the code to always use M2Crypto...

wottop:
WIll wait for a vdsm system with the proper certificate setup. Need to see the failure first.

kahowell:
673e700..53b77e8: rebased against master, added a commit that allows forcing M2Crypto usage via RHSM_USE_M2CRYPTO.

@wottop sending an email w/ testing instructions.

wottop:
ack

wottop (approved):

PR #141, opened 2018-05-02
1542652: When the -c option is used, don't parse the default files

nikosmoum:
- Now, when config files are passed through the cli -c option, configurations
in /etc/virt-who.conf and /etc/virt-who.d/* are not parsed.
- 'global' and 'default' sections are not parsed from files passed through
the -c option, but they will still be read from /etc/virt-who.conf.
- 'global' and 'default' sections are not parsed from /etc/virt-who.d/*

### DIFF
--- a/virtwho/config.py
+++ b/virtwho/config.py
@@ -1417,16 +1417,24 @@ def init_config(env_options, cli_options, config_dir=None):
         del effective_config[VW_ENV_CLI_SECTION_NAME]
 
     all_sections_to_add = {}
-    # read additional sections from /etc/virt-who.conf
-    all_sections_to_add.update(vw_conf)
-    # also read all sections in conf files in the drop dir
-    all_sections_to_add.update(effective_config.all_drop_dir_config_sections(config_dir=config_dir))
-    # We should ignore any additional sections defined as VW_GLOBAL as we've already parsed that
+
+    # read the files provided to the configs var if defined
+    for file_name in effective_config[VW_GLOBAL]['configs']:
+        logger.info("Using configuration passed in by -c/--configs; ignoring configuration files in '%s'", config_dir)
+        all_sections_to_add.update(parse_file(filename=file_name))
+
+    if len(all_sections_to_add) == 0:
+        # read sections from /etc/virt-who.conf
+        all_sections_to_add.update(vw_conf)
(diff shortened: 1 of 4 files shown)


virtwho/config.py
```diff
@@ -1417,17 +1417,24 @@ def init_config(env_options, cli_options, config_dir=None):
...
+
+    # read the file provided to the configs var if defined
     for file_name in effective_config[VW_GLOBAL]['configs']:
         all_sections_to_add.update(parse_file(filename=file_name))
 
+    if len(all_sections_to_add) == 0:
+        # read sections from /etc/virt-who.conf
+        all_sections_to_add.update(vw_conf)
+        # also read all sections in conf files in the drop dir
+        all_sections_to_add.update(effective_config.all_drop_dir_config_sections(config_dir=config_dir))
```
jirihnidek, line 1430: When configuration files from /etc/virt-who.d (it could be also some other directory) are not used, then this information should be printed to log file using `logger.info()` with explanation. I would prefer to print that all files from `config_dir` are ignored.  You can also list all ignored files in the `config_dir`. It's up to you.

virtwho/config.py
```diff
@@ -1417,17 +1417,24 @@ def init_config(env_options, cli_options, config_dir=None):
...
-    # read additional sections from /etc/virt-who.conf
-    all_sections_to_add.update(vw_conf)
-    # also read all sections in conf files in the drop dir
-    all_sections_to_add.update(effective_config.all_drop_dir_config_sections(config_dir=config_dir))
-    # We should ignore any additional sections defined as VW_GLOBAL as we've already parsed that
-    if VW_GLOBAL in all_sections_to_add:
-        del all_sections_to_add[VW_GLOBAL]
-    # also read the files in the configs list from the configs var if defined
+
+    # read the file provided to the configs var if defined
```
jirihnidek: The logic is little bit weird, but it's not your fault. Method `init_config` needs refactoring. :smile:

jirihnidek (changes requested):
Overall it's good PR. Please add some "info" message to log file, when -c/--config is used.

virtwho/config.py
nikosmoum: logging added

jirihnidek (approved):
ACK

PR #142, opened 2018-05-07
Fixing python 2 -> 3 issue

wottop:
import of ssl still needed
xmlrpc had a change of package name


cnsnyder (approved):
Looks good to me!

cnsnyder:
@wottop Any BZ associated with this? If not I'll merge as is.

wottop:
no BZ, just found it along the way

PR #144, opened 2018-05-22
1575513: Re-add changelog entries that were merged out.

wottop:


cnsnyder:
@wottop This looks good to me. We should get Igor's opinion as well to better understand his intentions.

My only question would be, should we change the URL and Source0 back to upstream fedora?

igor-raits:
this is not enough. you've re-added BuildRoot , %clean also.

wottop:
Removed build root and clean. Updated URL and Source0

kahowell:
@ignatenkobrain how does it look now? Also are there any documents which recommend practices for maintaining the spec file in upstream repository? Otherwise, I'd like to discuss what the best practices and processes for maintenance of virt-who and other candlepin-related projects in Fedora is (maybe such a doc could be produced as a result)?

igor-raits:
> how does it look now?

lgtm

> Also are there any documents which recommend practices for maintaining the spec file in upstream repository?

Recomendation is not to. Or you can do this, but anytime you push into fedora git, you need to check all changes and not overwrite them.

kahowell:
@ignatenkobrain ack. We will plan to modify our process to regularly incorporate changes from Fedora git into upstream.

PR #145, opened 2018-05-29
1455062: Partial fix of high CPU usage, when many conf files used

jirihnidek:
* Bug fix: https://bugzilla.redhat.com/show_bug.cgi?id=1455062
* Removed unused gc.collect(), because virt-who is not mulitprocess anymore.
* It is not necessary to do deepcopy twice. Deep copy is probably the
  biggest reason of high CPU load.
* Do not do deepcopy, when datastore is locked.
* Do not use exit for name of variable.
* Added delta_time for testing of granularity of sleeping.
* TODO: Do not use deepcopy, when host-to-quest mapping is added to
  datastore, but add there only necessary informations.

### DIFF
--- a/virtwho/datastore.py
+++ b/virtwho/datastore.py
@@ -33,21 +33,21 @@ class Datastore(object):
     def put(self, key, value):
         """
         Stores the value, retrievable by key, in a threadsafe manner in the
-        underlying datastore. (Assumes all items are pickleable)
+        underlying datastore.
 
         @param key: The unique identifier for this value
         @type  key: str
 
         @param value: The object to store
         """
+        to_store = copy.deepcopy(value)
         with self._datastore_lock:
-            to_store = copy.deepcopy(value)
             self._datastore[key] = to_store
 
     def get(self, key, default=None):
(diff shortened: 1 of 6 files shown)


virtwho/datastore.py
```diff
@@ -60,8 +60,9 @@ def get(self, key, default=None):
         """
         with self._datastore_lock:
             try:
-                item = copy.deepcopy(self._datastore[key])
-                return item
+                # item = copy.deepcopy(self._datastore[key])
+                # return item
```
wottop: remove comments

wottop (commented):
Where is the CPU usage specifically relieved. Is it the adjustment to the wait counters?

wottop:
Please show in the commit information that the changing of deep copy use is for the CPU usage issue. With several changes made, some not related to the bug, it will make it easier for someone reading through the commit messages later.

jirihnidek:
@wottop I updated commit message and first comment of this PR according commit message.

wottop:
ack

PR #149, opened 2018-06-13
ENT-554 Host reports for libvirt and rhevm include the system hardware uuid

wottop:
The fact is added for scenarios where there can be a regular consumer check in
 and hypervisor reporting on the same host system.

Related BZ's: 1314902, 1410601, 1566000, 1576110


jirihnidek (changes requested):
Overall it looks good. I added one unit test for libvirtd. Please add more description to this PR/commit.

I can see that this PR is part of ENT-554 (please add it to label of PR and commit). Is it related to any BZ bug report?

jirihnidek (approved):
ACK

PR #150, opened 2018-06-14
Correction to the spec file condition for python 3

wottop:
Review based on Rel-Eng JIRA response.

https://projects.engineering.redhat.com/browse/RCM-35339


cnsnyder (approved):
Looks good to me. I would try it first however!

PR #151, opened 2018-06-18
1368341: Warn that --sam/--satellite6 are unused & deprecated ENT-251

nikosmoum:
* Now logging a warning when --sam/--satellite6 are used.
* man page and --help output updated to explaing that these options are unused and virt-who will report to either sam/satellite/stage candlepin regardless of their being there.

### DIFF
--- a/virt-who.8
+++ b/virt-who.8
@@ -50,13 +50,13 @@ Choose virtualization backend that should be used to gather host/guest associati
 Choose where the host/guest associations should be reported
 .TP
 \fB\-\-sam\fR
-[Deprecated] Report host/guest associations to the Subscription Asset Manager or Satellite 6 [default]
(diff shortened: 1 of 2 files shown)


virt-who.8
```diff
@@ -50,13 +50,13 @@ Choose virtualization backend that should be used to gather host/guest associati
 Choose where the host/guest associations should be reported
 .TP
 \fB\-\-sam\fR
-[Deprecated] Report host/guest associations to the Subscription Asset Manager or Satellite 6 [default]
+[Deprecated] Report host/guest associations to the Subscription Asset Manager, Satellite 6, or Stage Candlepin. This option specifies the default behaviour, and thus it is not used [default]
```
wottop: 'Stage Candlepin' is not what the end user should see. Let's use 'Red Hat Subscription Management (RHSM)'

wottop (changes requested):

virt-who.8
nikosmoum: Updated

wottop (approved):

wottop:
ack

PR #154, opened 2018-06-25
Update for build process

wottop:
Needs Distirubution Buildler so that the tar ball is created correctly for the RPM tool.
Uncomment on the placeholder bug in 7.6


kahowell (approved):
Ack

PR #156, opened 2018-06-27
1510920: Change the choreography for the job status check

wottop:
The order has been changed so that the status is not checked directly
 after the submission. The current method was resulting in multiple
 status check retries while the checkin was being processed.

This version will check immediately before the next run. If the previous
 job has not yet been processed, it will be bypassed until the next interval

One shot reports will no longer wait for the status to be returned.

### DIFF
--- a/virtwho/manager/subscriptionmanager/subscriptionmanager.py
+++ b/virtwho/manager/subscriptionmanager/subscriptionmanager.py
@@ -280,7 +280,7 @@ class SubscriptionManager(Manager):
             raise ManagerError("Communication with subscription manager failed: %s" % str(e))
 
         if is_async is True:
-            report.state = AbstractVirtReport.STATE_PROCESSING
+            report.state = AbstractVirtReport.STATE_CREATED
             report.job_id = result['id']
         else:
             report.state = AbstractVirtReport.STATE_FINISHED
--- a/virtwho/virt/virt.py
+++ b/virtwho/virt/virt.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 from __future__ import print_function
+
+
 """
 Module for abstraction of all virtualization backends, part of virt-who
 
@@ -482,6 +484,7 @@ class DestinationThread(IntervalThread):
         self.is_initial_run = True
         self.source_keys = source_keys
         self.last_report_for_source = {}  # Source_key to hash of last report
+        self.submitted_report_and_hash_for_source = {}  # Source key to submitted batch report and hash
         self.options = options
         self.reports_to_print = []  # A list of reports we would send but are
         #  going to print instead, to be used by the owner of the thread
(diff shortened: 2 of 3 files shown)


virtwho/manager/subscriptionmanager/subscriptionmanager.py
```diff
@@ -369,6 +369,22 @@ def check_report_state(self, report):
             self.logger.debug("Number of mappings unchanged: %d", len(result_data.get('unchanged', [])))
             self.logger.info("Mapping for config \"%s\" updated", report.config.name)
 
+    def cancel_job(self, report):
```
jirihnidek: This function isn't used.

virtwho/virt/virt.py
```diff
@@ -1,5 +1,8 @@
 # -*- coding: utf-8 -*-
 from __future__ import print_function
+
+from rhsm.connection import RestlibException
```
jirihnidek: It seems that imported class `RestlibException` is not used.

virtwho/virt/virt.py
```diff
@@ -756,6 +732,38 @@ def _send_data(self, data_to_send):
...
+                    break
+                retry_after = self.handle_429(e.retry_after, num_429_received)
+                self.logger.debug('429 encountered while checking job '
+                                  'state, checking again in "%s"', retry_after)
+                self.interval_modifier = retry_after
+            except (ManagerError, ManagerFatalError):
+                self.logger.exception("Error during job check: ")
+                break
+            # If we get here and have to try again, it's not our first rodeo...
+            first_attempt = False
```
jirihnidek, line 769: Nitpick: there should be two empty lines between two classes.

virtwho/virt/virt.py
```diff
@@ -756,6 +732,38 @@ def _send_data(self, data_to_send):
                                 if source_key not in sources_sent]
         return
 
+    def check_report_status(self, report):
```
jirihnidek, line 734: Please add some doc string for this method.

virt-who.spec
```diff
@@ -37,9 +37,9 @@ Requires:       %{python_ver}-setuptools
 # python-rhsm 1.20 has the M2Crypto wrappers needed to replace M2Crypto
 # with the python standard libraries where plausible
 %if %{use_python3}
-Requires:       python3-subscription-manager-rhsm
+Requires:       python3-subscription-manager-rhsm >= 1.21.6
 %else
-Requires:       subscription-manager-rhsm
+Requires:       subscription-manager-rhsm >= 1.21.6
```
jirihnidek: BTW: Why is required this version?

jirihnidek (changes requested):
Overall, it is good PR. I only request some small changes.

virtwho/manager/subscriptionmanager/subscriptionmanager.py
wottop: it was an earlier idea where the job would get canceled if not finished. I had left it in in case we ever need to make that cancel happen.

virt-who.spec
wottop: See above comment on cancel_job.  This is the version that has the cancel_job call in it.

jirihnidek (approved):
ACK

PR #157, opened 2018-06-29
ENT-493: Add option to command line to return version

vritant:


jirihnidek (changes requested):
Please fix unit tests first and add new unit test for `--version` option. Hint: function `parse_options` returns more items in tuple.

virtwho/parser.py
```diff
@@ -305,7 +305,7 @@ def parse_cli_arguments():
     """
     parser = ArgumentParser(
         usage="virt-who [-d] [-i INTERVAL] [-o] [--sam|--satellite5|--satellite6] "
-              "[--libvirt|--vdsm|--esx|--rhevm|--hyperv|--xen|--kubevirt]",
+              "[--libvirt|--vdsm|--esx|--rhevm|--hyperv|--xen|--kubevirt|--version]",
```
jirihnidek: It seems like --version is another type of hypervisor backend. It's not true.  Add there something like `[--help] [--version]`

virtwho/main.py
```diff
@@ -109,30 +109,40 @@ def reload(signal, stackframe):
...
+        logger, effective_config, cli_options = parse_options()
         # We now have the effective_config
     except OptionError as e:
         print(str(e), file=sys.stderr)
         exit(1, status="virt-who can't be started: %s" % str(e))
 
+    if 'version' in cli_options and cli_options['version']:
+        print(get_version())
+        exit(os.EX_OK)
+
```
jirihnidek: What about this approach? Move this block of code to `parser.py` (line 486). And move method `get_version()` to `parser.py`. It will simplify `parse_options()` method. It will not be necessary to return `cli_options` to `main.py`. Another benefit is following: it will not print all errors and warnings, when `--version` is used. It will only print version to stdout.

jirihnidek (changes requested):
I believe it is possible to simplify code and get better results. Loot at my comments.

virtwho/main.py
vritant: @jirihnidek thats actually how I implemented it first, but that did not seem like the right thing to do, a parser actually performing an action. 

anywho, if you are up for it, iv made the suggested changes

jirihnidek (approved):
ACK

PR #160, opened 2018-07-13
1581021: Decode error from unicode passwords

wottop:
Updated username/password read from config to unicode

Related 1581022: username/password sent through BasicAuth as bytes

### DIFF
--- a/tests/test_hyperv.py
+++ b/tests/test_hyperv.py
@@ -1,3 +1,5 @@
+# -*- coding: utf-8 -*-
+
 from __future__ import print_function
 """
 Test of Hyper-V virtualization backend.
@@ -141,7 +143,7 @@ class TestHyperV(TestBase):
             'type': 'hyperv',
             'server': 'localhost',
             'username': 'username',
-            'password': 'password',
+            'password': u'1€345678',
             'owner': 'owner',
             'env': 'env,'
         }
--- a/tests/test_rhevm.py
+++ b/tests/test_rhevm.py
@@ -1,3 +1,5 @@
+# -*- coding: utf-8 -*-
+
 from __future__ import print_function
 """
 Test of RHEV-M virtualization backend.
@@ -104,7 +106,7 @@ class TestRhevM(TestBase):
 
     def setUp(self):
         config = self.create_config(name='test', wrapper=None, type='rhevm', server='localhost', username='username',
-                        password='password', owner='owner', env='env')
+                        password=u'1€345678', owner='owner', env='env')
         self.rhevm = Virt.from_config(self.logger, config, Datastore())
         self.rhevm.major_version = '3'
         self.rhevm.build_urls()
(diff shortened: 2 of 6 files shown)


jirihnidek (changes requested):
Hi, Good news is that it seems that Hyper-V probably works as expected. Bad news is, that RHEVM doesn't work on RHEL7. RHEL7 has Python module `request` in version 2.6.0 and this version tries to do encoding to `latin1`. See following error:

```
2018-07-24 09:46:30,744 [virtwho.main ERROR] MainProcess(4280):Thread-2 @virt.py:run:396 - Thread 'sat-rhevm' fails with exception:
  File "virtwho/virt/virt.py", line 387, in run
    self._run()
  File "virtwho/virt/virt.py", line 339, in _run
    self.prepare()
  File "virtwho/virt/rhevm/rhevm.py", line 121, in prepare
    self.get_version()
  File "virtwho/virt/rhevm/rhevm.py", line 148, in get_version
    verify=False)
  File "/usr/lib/python2.7/site-packages/requests/api.py", line 68, in get
    return request('get', url, **kwargs)
  File "/usr/lib/python2.7/site-packages/requests/api.py", line 50, in request
    response = session.request(method=method, url=url, **kwargs)
  File "/usr/lib/python2.7/site-packages/requests/sessions.py", line 450, in request
    prep = self.prepare_request(req)
  File "/usr/lib/python2.7/site-packages/requests/sessions.py", line 381, in prepare_request
    hooks=merge_hooks(request.hooks, self.hooks),
  File "/usr/lib/python2.7/site-packages/requests/models.py", line 308, in prepare
    self.prepare_auth(auth, url)
  File "/usr/lib/python2.7/site-packages/requests/models.py", line 495, in prepare_auth
    r = auth(self)
  File "/usr/lib/python2.7/site-packages/requests/auth.py", line 50, in __call__
    r.headers['Authorization'] = _basic_auth_str(self.username, self.password)
  File "/usr/lib/python2.7/site-packages/requests/auth.py", line 30, in _basic_auth_str
    b64encode(('%s:%s' % (username, password)).encode('latin1')).strip()
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 21: ordinal not in range(128)
```

Fedora 27 has `request` module in version 2.8.14 and this version doesn't try to encode password nor username to latin-1 encoding. BTW: The reason, why request module (version 2.8.14) does not try to do encoding to latin-1 is mystery for me (see method `_basic_auth_str` in file `/usr/lib/python2.7/site-packages/requests/auth.py`).

Can we say that virt-who has to use password including only ASCII characters on RHEL7?

jirihnidek:
BTW: I discovered, why this Python code returns `False`:

```python
password = 'redhat€'
type(password)
# str is returned as expected
isinstance(password, str)
# False is returned as not expected
```

The reason is following. Some module probably uses following on the top of file:

```python
from __future__ import unicode_literals
```

wottop:
Yes. We must make that constraint.

This fix will take care of hyperv and will fix rhevm for everything except RHEL 7 and older.
Fedora should be fixed for newer python-requests versions.

Please judge PR based on that. Thanks.

jirihnidek (approved):
ACK

PR #165, opened 2018-08-13
1577954: Added config option filter_type; ENT-580

jirihnidek:
* BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1577954
* Added new config file option filter_type. Value of this option
  can be one of: 'wildcards, 'regex'.
* When filtering is used and filter_type is not specified, then
  warning is printed to log file. Behavior should be the same.

### DIFF
--- a/tests/test_config_section_virt.py
+++ b/tests/test_config_section_virt.py
@@ -42,7 +42,8 @@ LIBVIRT_SECTION_VALUES = {
     'env': '123456',
     'owner': '123456',
     'hypervisor_id': 'uuid',
-    'filter_hosts': '*.example.com'
+    'filter_hosts': '*.example.com',
+    'filter_type': 'wildcards'
 }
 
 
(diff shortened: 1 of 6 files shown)


tests/test_config_section_virt.py
```diff
@@ -275,6 +276,33 @@ def test_validate_filter(self):
...
+    def test_validate_regex_filter_type(self):
+        """
+        Test validation of regex filter type
+        """
+        self.init_virt_config_section()
+        self.virt_config['filter_type'] = 'regex'
+        result = self.virt_config._validate_filter_type('filter_type')
+        self.assertIsNone(result)
+
     def test_validate_filter_hypervisor_id_hostname(self):
```
wottop, line 306: Is there a test for a regex and no filter type set?
I'd like to see a test of this fall back.

wottop (changes requested):

tests/test_config_section_virt.py
jirihnidek: @wottop This fallback is covered by existing unit tests in `test_virt.py'. I added more tests there and I also updated manual page of `virt-who-config.5`.

wottop:
go ahead and squash. ack otherwise.

jirihnidek:
@wottop commits are squashed

wottop:
ack

PR #167, opened 2018-08-17
1387800: set name of ESX cluster properly

jirihnidek:
* BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1387800#c17
* It wasn't so obvious that name of cluster displayed in vSphere UI
  is something else than we can see as ID of cluster in XML.
* Fixed coresponding unit tests.


wottop:
ack

PR #169, opened 2018-08-29
Print/log debug information about filtered hosts

jirihnidek:
* Debug printing/logginng of filtered hosts did not work at all
* When host passes filtering, then debug print is also printed.


wottop:
ack

PR #172, opened 2018-08-31
kubevirt: Ignore vmis in Scheduling

pkliczewski:
There is time when virtual machine instance is not assigned to specific
host but it appears in vmi list. Such vmis should be reported only when
scheduled on specific host.


pkliczewski:
@wottop please review

wottop:
ack

PR #173, opened 2018-08-31
1557296: Ignore commented out lines prefixed with space/tab (ENT-606)

nikosmoum:


jirihnidek (changes requested):
I don't think that overriding of `_read()` method is good idea. If there would be any future bug fix provided in `configparser` module, then we will still use old and obsolete code.

Your code also breaks specification of ini file. Let's imagine somebody is using very long (generated) password broken in several lines containing `#` at beginning of line like this:

```
password = "Iu*,qo)yqmsP@^
    a@lkd$jFn4kj32
    #,qoa;(0jWjnk$k
    r%kj9n30l2k3TjG
    "
```

This is completely correct configuration file, but your PR removes part of this password.

I would print only warnings when there is `%0A%23` using something like this:

```python
def _read(self, fp, fpname):
    while line = fp.readline()
        # detect %0A%23 and print warning
    fp.seek(0)
    SafeConfigParser._read(self, fp, fpname)
```

nikosmoum:
@jirihnidek fixed

nikosmoum:
@jirihnidek I just realized by trying to fix the failing python3 test, that in python3, the configparser lib will ignore line continuations if they start with '#' anyway. (so this problem only occurs with python2)

jirihnidek (approved):
ACK

PR #176, opened 2018-09-07
1369634: Dont log proxy html errors for hyperv: (ENT-613)

nikosmoum:
- When hyperv gets an HTML page as error response from a proxy, don't log the whole html, but try to scrape the title off of it.
- If scraping the title doesn't work, only log the http error code.

### DIFF
--- a/virtwho/virt/hyperv/hyperv.py
+++ b/virtwho/virt/hyperv/hyperv.py
@@ -390,21 +390,23 @@ class HyperVSoap(object):
         else:
             data = response.content
             try:
-                xml = ElementTree.fromstring(data)
-                errorcode = xml.find('.//{http://schemas.dmtf.org/wbem/wscim/1/cim-schema/2/MSFT_WmiError}error_Code')
+                xml_doc = ElementTree.fromstring(data)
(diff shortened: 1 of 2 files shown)


virtwho/virt/hyperv/hyperv.py
```diff
@@ -395,9 +395,11 @@ def post(self, body):
                 # Suppress reporting of invalid namespace, because we're testing
                 # both old and new namespaces that HyperV uses
                 if errorcode is None or errorcode.text != '2150858778':
-                    self.logger.debug("Invalid response (%d) from Hyper-V: %s", response.status_code, data)
+                    title = xml.find('.//title')
```
jirihnidek: I know that you are using existing variable and it's not your fault, but please do small refactoring here and rename `xml` to something like `xml_doc`, because `xml` is well known Python module (`ElementTree` is imported from this module). It could cause some issues in the future.

jirihnidek (changes requested):
Overall it is good PR. Please do small refactoring of code to avoid future problems.

nikosmoum:
@jirihnidek fixed all variables in this file called `xml`

jirihnidek (approved):
ACK

PR #178, opened 2018-09-18
Install subscription-manager, not python-rhsm which is deprecated:

nikosmoum:
- Removed python-rhsm from the requirements, and added subscription-manager as dependency.
- Added some dependencies that travis requires to install subscription-manager.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,12 @@ env:
 sudo: required
 dist: trusty
 install:
+  - pip install -U setuptools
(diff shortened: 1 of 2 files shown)


nikosmoum:
Note: this PR should be merged only after https://github.com/candlepin/subscription-manager/pull/1920 has been merged, since installing subman as a requirement in ubuntu will work only with the changes in that PR.

wottop:
ack

PR #181, opened 2018-10-30
1637407: vCenter mapping info failure due to TypeError

wottop:
Failure occured when running aginst python 3.

To see the fix run, test_esx.py on python 3 in master then on this branch.


jirihnidek (approved):
I can confirm that it works in Python2 and Python3. Thanks for PR.

PR #182, opened 2018-10-31
ENT-896: Disable deprecated configuration options in python 3

wottop:


jirihnidek (approved):
ACK. Thanks for the PR.

PR #184, opened 2018-11-05
1638250: Proxy issue when https not specified

wottop:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1638250


jirihnidek (approved):
ACK

PR #187, opened 2018-11-30
fix travis setup of packages

wottop:


wottop:
By using a dist value that does not correspond to any recognized one [I used "current"], it appears that the most recent one will be used as a fall back.

kahowell (approved):

PR #188, opened 2018-12-06
Fixing kubevirt config path argument

pkliczewski:
We expected to get location of kubernetes config file provided as env
variable and there was no way to configure it only in config file. This
PR moves this setting to config file.

Fixes: https://bugzilla.redhat.com/1654652


pkliczewski:
@wottop please review

pkliczewski:
@wottop can you please take a look?

wottop:
We are working to remove environment variables from the virt-who process. Is there a way to use the virt-who.conf file it achieve this functionality?

pkliczewski:
@wottop This patch enables usage of config file. Before this change we were dependent on environment variable.

virt-who.conf
```diff
@@ -90,5 +90,5 @@ VIRTWHO_DEBUG=0
 # Options for Kubevirt mode
 #VIRTWHO_KUBEVIRT_OWNER=
 #VIRTWHO_KUBEVIRT_ENV=
-#KUBECONFIG=
+#VIRTWHO_KUBEVIRT_KUBECONFIG=
```
wottop, line 93: This is the name of an environment variable. Shouldn't it be 'kubeconfig'?

wottop (changes requested):

virt-who.conf
pkliczewski: We are not using environment variable anymore. We rely on command line or config file now.

wottop: then remove this line from this file.

wottop: Are you aware that you can specify a single config file on the command line, using '-c'?

pkliczewski: I did not know this option. Thanks for a hint.

wottop: Does that reduce the necessity of this change?

pkliczewski: No, currently kubevirt plugin won't use this property from config file at all. This PR makes it possible.

wottop:
ack

wottop (approved):

pkliczewski:
:tada:

PR #189, opened 2018-12-12
1657104: Remove references to removed command line options

wottop:


nikosmoum (changes requested):
One small change before merging: The Synopsis part in the man page still includes the deprecated options, so they should probably also be removed from [there](https://github.com/candlepin/virt-who/pull/189/files#diff-61447b64588f1eaddd4694907ea46419R5).

nikosmoum (approved):
ACK

PR #190, opened 2018-12-21
Update releaser

wottop:


barnabycourt:
What is this updating the releaser to do?

wottop:
F27 builds are no longer allowed
The flag to check for 8.0 is just the 'release+'

jirihnidek (approved):
ACK

PR #193, opened 2019-02-21
1486270: pass no_proxy from config files

wottop:
For testing, the parameter can be placed in either the /etc/virt-who.conf file or the conf file you are using in /etc/virt-who.d

this is the config file I have been using:

[stage-libvirt]
type=libvirt
owner=711497
env=Library
server=qemu:///
#server=10.66.129.67
username=root
password=test
rhsm_proxy_hostname = ktordeur-proxy.usersys.redhat.com
rhsm_proxy_port = 3128
rhsm_proxy_user = proxy
rhsm_proxy_password = proxy
#rhsm_no_proxy =  *


jirihnidek (changes requested):
The smoke test works as expected. Code is OK too. Please change only one small thing. Commit message should start with 1486270 (not 148670).

PR #195, opened 2019-03-02
Nutanix AHV support

mrjoshuap:
See Bugzilla [1664339](https://bugzilla.redhat.com/show_bug.cgi?id=1664339)

I've built out tentative support for Nutanix AHV.  This does not utilize fakevirt and instead utilizes Nutanix API v2.  I've not performed extensive testing, but the mappings are returned correctly.

I also built out the same/similar tests as seen in RHEVM and ESX.

[All tests are currently passing](https://travis-ci.org/mrjoshuap/virt-who/builds/500906761)

Please advise.


mrjoshuap:
TODO: Pagination of results

aeibagi:
Hi @mrjoshuap 

Thanks for this effort of having Nutanix support for virt-who. We are currently working on this fix in-house (at Nutanix) which also includes adding the support for PC which uses V3 API. We would like to collaborate more with you regarding the solution.

Thanks

mrjoshuap:
Thanks @aeibagi for the offer.

I stubbed together some outputs from the v2 api and didn't get to the v3 api before my 2 hours ended.  Functionally speaking, if you have some good high level docs that describe the differences that'd be great.

I also don't have a easy access to test it for real, though I made sure it had some unit tests.

Any chance you could provide v3 outputs from the following v2 api calls?

vm
vms
host
hosts
cluster
clusters

I'm still requiring pagination, but haven't had enough large enough datasets to test with, so any outputs in that regard would be greatly appreciated.

I could get v3 stubbed out without connectivity with moderate ease with those bits.

Thanks!!!

mrjoshuap:
One other thing that I haven't been able to verify, but maybe you can confirm.

Does a VMs hardware UUID returned in the API calls equal the output of dmidecode?

`dmidecode |grep -i uuid | tr '[:upper:]' '[:lower:]'`

I've not been able to confirm, but if that is not a true statement, it will require a major rethink in the approach of this backend.

aeibagi:
@mrjoshuap 

We Have fully implemented a solution with works with v2 and v3 API. It has passed the initial testing with the real setup. It’s compatible with both Prism central and prism element. We will be pushing the solution into Nutanix Open source community very soon if you wish to use that solution.

Also If you wish to accept our implementation then we can integrate our solution here so we avoid the duplicate effort.

To Answer your questions:

V3 Api output has 3 main keys:
‘Status’, ‘Spec’ and ‘Metadata’

Here is the link for the v3 lists/clusters, you can find other entity calls from this page:
https://developer.nutanix.com/reference/prism_central/v3/api/clusters/postclusterslist



Also, 

Does a VMs hardware UUID returned in the API calls equal the output of dmidecode?
Yes

wottop:
Is this still a viable pr? there is  another one on the list that is not in progress.

wottop:
The the other PR has been merged which makes this obsolete. Closing the PR.

PR #199, opened 2019-03-28
1693858: Send hardware uuid on every check in for system reconcilliation

wottop:
Added the constant sending of dmi.system.uuid for additional fabrics
esx
xen
hyperv


jirihnidek (approved):
ACK

PR #200, opened 2019-03-29
1667522: Omit ESX host from report when no hostname is present

wottop:
The lack of hostname is because the host is off. It will cause problems down the line if an
 empty hostname is sent to the server.

### DIFF
--- a/tests/test_esx.py
+++ b/tests/test_esx.py
@@ -166,6 +166,41 @@ class TestEsx(TestBase):
         self.assertEqual(expected_result.toDict(), result.toDict())
 
     @patch('suds.client.Client')
+    def test_getHostGuestMappingNoHostName(self, mock_client):
+        expected_hypervisorId = 'Fake_uuid'
+        expected_guestId = 'guest1UUID'
+        expected_guest_state = Guest.STATE_RUNNING
+
+        fake_parent = MagicMock()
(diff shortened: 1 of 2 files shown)


tests/test_esx.py
```diff
@@ -165,6 +165,41 @@ def test_getHostGuestMapping(self, mock_client):
...
+
+        fake_vm_id = MagicMock()
+        fake_vm_id.value = 'guest1'
+
+        fake_vm = MagicMock()
+        fake_vm.ManagedObjectReference = [fake_vm_id]
+        fake_vms = {'guest1': {'runtime.powerState': 'poweredOn',
+                               'config.uuid': expected_guestId}}
+        self.esx.vms = fake_vms
+
```
jirihnidek, line 186: Nitpick: I would add here note that: `'config.network.dnsConfig.hostName'` is not listed in `fake_host`.

jirihnidek (approved):
ACK.

tests/test_esx.py
wottop: That is the point of the test. The hypervisor does not get reported on when the hostname is absent.

jirihnidek: Yeah, I know. :-) It would be nice to have such information in comment in source code, but comment here is enough too. :-)

PR #201, opened 2019-04-05
1695519: use correct uuid in kubevirt report

pkliczewski:
We used 'namespace/name' as guest_id which is not correct. This PR
changes it to use firmware uuid instead.


pkliczewski:
@wottop please take a look

wottop:
Our normal procedure is to prepend the commit message with the bugzilla number where applicable.

i.e. 1695519: use correct uuid in kubevirt report

This allows for our build script to associate bz's

pkliczewski:
@wottop Both PR name and commit message updated

wottop:
ack

PR #202, opened 2019-04-05
1695538: Provide support for hypervisor_id option

pkliczewski:
We have 3 values supported for hypervisor_id but due to Kubernetes limitations
we can only support uuid and hostname. This PR adds support for both values.

### DIFF
--- a/virtwho/virt/kubevirt/kubevirt.py
+++ b/virtwho/virt/kubevirt/kubevirt.py
@@ -29,6 +29,7 @@ from virtwho.virt.kubevirt.client import KubeClient
 class KubevirtConfigSection(VirtConfigSection):
 
     VIRT_TYPE = 'kubevirt'
+    HYPERVISOR_ID = ('uuid', 'hostname')
 
     def __init__(self, section_name, wrapper, *args, **kwargs):
         super(KubevirtConfigSection, self).__init__(section_name,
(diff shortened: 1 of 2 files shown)


pkliczewski:
@wottop please review

wottop:
needs rebasing

virtwho/virt/kubevirt/kubevirt.py
```diff
@@ -84,14 +85,23 @@ def getHostGuestMapping(self):
             status = node['status']
             version = status['nodeInfo']['kubeletVersion']
             name = node['metadata']['name']
-            host_id = status['nodeInfo']['machineID']
-            address = status['addresses'][0]['address']
+
+            uuid = status['nodeInfo']['machineID']
```
wottop, line 89: This should be the system uuid. It need to be an immutable and unique identifier for the host system. If it is, then this is OK.

wottop (changes requested):

virtwho/virt/kubevirt/kubevirt.py
pkliczewski: Kubernetes guys suggest [1] to use `machineID` instead.

> SystemUUID reported by the node. For unique machine identification MachineID is preferred

[1] https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/#nodesysteminfo-v1-core

wottop: For this purpose, yes.

PR #204, opened 2019-04-25
1506167: Ignore new SIGHUP signals during signal handling

jirihnidek:
* BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1506167
* This bug was Python 2 specific.
* When signal SIGHUP was handled, then recieving other signal
  caused terminating of virt-who


wottop:
Needs a better commit message:

"Ignore new SIGHUP signals during signal handling"

wottop:
Otherwise, ack

wottop:
ack

PR #205, opened 2019-05-06
1641953: Virt-who fails if one hypervisor has wrong encrypted password

wottop:
In python3 an error can be thrown on incorrect encrypted password that was not previously accounted for.

### DIFF
--- a/tests/test_config_section_virt.py
+++ b/tests/test_config_section_virt.py
@@ -352,3 +352,20 @@ class TestVirtConfigSection(TestBase):
         ]
         result = self.virt_config._validate_filter('filter_hosts')
         self.assertIsNotNone(result)
+
+    def test_validate_wrong_values_in_encrypted_password(self):
+        """
+        Test of validation of corrupted encrypted password
+        """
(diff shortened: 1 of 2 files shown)


jirihnidek (dismissed):
Overall it is good change. Please add following unit test to file: `./tests/test_config_section_virt.py`:

```python
    def test_validate_wrong_values_in_encrypted_password(self):
        """
        Test of validation of corrupted encrypted password
        """
        self.mock_pwd_file()
        # Safe current password
        password = self.virt_config['password']
        # Delete unencrypted password first
        del self.virt_config['password']
        # Set up corrupted encrypted password
        encrypted_pwd = Password.encrypt(password)
        corrupted_encrypted_pwd = encrypted_pwd + 'xxxx'
        self.virt_config['encrypted_password'] = corrupted_encrypted_pwd
        # Do own testing here
        result = self.virt_config._validate_encrypted_password('encrypted_password')
        self.assertIsNotNone(result)
        self.unmock_pwd_file()
```

PR #209, opened 2019-05-16
Add Nutanix AHV support for virt-who

aeibagi:
This solution is no longer using the fake virt. It uses both v2 and v3 api. When we are only dealing with a PE (prism element or rather a cluster) the solution utilizes the v2 api and builds the map. If a prism central is present and defined in the config file, it then uses v3 api to retrieve the information for the clusters register to the prism element and build the map. 
This solution also listens and captures any vm operation events and then automatically build the new map if there are any new changes. 

I've tested the solution against the actual setup using both v2 and v3.

### DIFF
--- a/tests/test_ahv.py
+++ b/tests/test_ahv.py
@@ -1,0 +1,562 @@
+from __future__ import print_function
+
+import six
+
+from base import TestBase
+from mock import patch, call, ANY, MagicMock
+from requests import Session
+from six.moves.queue import Queue
+from threading import Event
+
+from virtwho import DefaultInterval
+from virtwho.datastore import Datastore
+from virtwho.virt.ahv.ahv import AhvConfigSection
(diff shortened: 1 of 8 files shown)


wottop:
please correct the failure in the travis run

aeibagi:
Done, all checks passed:
https://travis-ci.org/candlepin/virt-who/builds/534929389?utm_source=github_status&utm_medium=notification

wottop:
Looks good. Please squash to a single commit.

aeibagi:
Done, all squashed into 1 commit

wottop:
Please rebase to master for the recent changes. Note that the environment variable has been purged [is not used by the server] and there were deprecated entries left behind in zsh.

aeibagi:
Rebased to the latest master and removed the env

wottop:
ack

PR #211, opened 2019-05-16
1516209: Configuration should be deemed invalid when server is not sp…

wottop:
…ecified


cnsnyder:
Thank you, this seems to work well.

PR #213, opened 2019-05-20
1530290: Remove enviroment as an input variable for the hypervisor ch…

wottop:
…eck in

The current server is not using this value so there is little need to send it


tests/test_config_section_libvirtd.py
```diff
@@ -249,10 +227,6 @@ def test_validate_sat5_env(self):
         self.virt_config['sm_type'] = 'satellite'
         # When server is set, then owner has to be set too
         assert 'server' in self.virt_config
```
jirihnidek: I would remove this method, because it doesn't test almost anything.

tests/test_fake.py
```diff
@@ -118,7 +117,6 @@ def test_read_hypervisor_from_non_hypervisor(self):
 type=fake
 is_hypervisor=true
 owner=covfefe
-env=covfefe
```
jirihnidek, line 121: :-)

jirihnidek (changes requested):
Thanks for PR. Overall It is good PR. I think it is necessary to remove `env` from man pages `virt-who-config.5`. BTW: there are also some obsolete CLI arguments in virt-who.zsh.

wottop:
Will squash after second revue

jirihnidek (approved):
ACK. Please squash and then it is ready to be merged to master.

wottop:
squashed

PR #214, opened 2019-05-21
1522384: Log at debug level when config uses default entries

wottop:
print_ option should added to virt-who.conf template


jirihnidek (approved):
ACK

wottop:
also fixed https://bugzilla.redhat.com/show_bug.cgi?id=1510712

PR #215, opened 2019-05-22
1523482: Interval value set to empty string will revert to default value

wottop:
Also goes for any non-integer string


jirihnidek (approved):
ACK

PR #216, opened 2019-05-28
Exclude AHV from RPM

wottop:


wottop:
Will deploy on the downstream instead

PR #217, opened 2019-05-29
1714133: can't set data property

pkliczewski:
There is a difference how data property is handled in py2 and py3.
We can't set the value when using py3.

Fixes: https://bugzilla.redhat.com/1714133


pkliczewski:
@wottop Please review

wottop:
ack

PR #218, opened 2019-05-30
1652549: Add heartbeat call to virt-who cycle

wottop:
This ensures that hypervisors will have updated lastcheckin times even
 if they show no changes on the hypervisor checkins.

### DIFF
--- a/tests/test_subscriptionmanager.py
+++ b/tests/test_subscriptionmanager.py
@@ -107,6 +107,20 @@ class TestSubscriptionManager(TestBase):
         self.sm.connection.return_value.has_capability = MagicMock(return_value=False)
 
     @patch('rhsm.connection.UEPConnection')
+    # def test_hypervisorHeartbeat(self):
+    def test_hypervisorHeartbeat(self, rhsmconnection):
+        owner = 'owner'
+        config = VirtConfigSection.from_dict({'type': 'libvirt', 'owner': owner}, 'test', None)
+        # Ensure we try out the new API
+        rhsmconnection.return_value.has_capability.return_value = True
+        self.sm.logger = MagicMock()
+        self.sm.hypervisorHeartbeat(config, options=None)
+        self.sm.connection.hypervisorHeartbeat.assert_called_with(
+            owner,
+            None
+        )
+
+    @patch('rhsm.connection.UEPConnection')
     def test_job_status(self, rhsmconnection):
         rhsmconnection.return_value.has_capability.return_value = True
         config = VirtConfigSection.from_dict({'type': 'libvirt', 'owner': 'owner'}, 'test', None)
--- a/virtwho/virt/virt.py
+++ b/virtwho/virt/virt.py
@@ -596,6 +596,11 @@ class DestinationThread(IntervalThread):
         @param data_to_send: A dict of source_keys, report
         @type: dict
         """
+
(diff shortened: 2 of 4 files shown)


wottop:
To test you will need to use this version of subscription-manager:
https://github.com/candlepin/subscription-manager/pull/2096

You will also need to build this with 'sudo make DESTDIR=/usr install' as the version requirement cannot be fulfilled until subman is tagged.

wottop:
I believe the subscription-manager PR needs to be merged before the travis test will work properly for this one. Please merge that one first.

tests/test_subscriptionmanager.py
```diff
@@ -106,6 +106,21 @@ def test_hypervisorCheckInAsync(self, rhsmconnection):
...
+        config = VirtConfigSection.from_dict({'type': 'libvirt', 'owner': owner}, 'test', None)
+        # Ensure we try out the new API
+        rhsmconnection.return_value.has_capability.return_value = True
+        self.sm.logger = MagicMock()
+        self.sm.hypervisorHeartbeat(config, options=None)
+        self.sm.connection.hypervisorHeartbeat.assert_called_with(
+            owner,
+            None
+        )
+        self.sm.connection.return_value.has_capability = MagicMock(return_value=False)
```
jirihnidek: Setting `has_capability` in mock object here has now effect.

virtwho/virt/virt.py
```diff
@@ -596,6 +596,12 @@ def _send_data(self, data_to_send):
         @param data_to_send: A dict of source_keys, report
         @type: dict
         """
+
+        if not isinstance(data_to_send, ErrorReport):
+            # Ping all hypervisors for this reporter to update check in time regardless of change
```
jirihnidek, line 601: I would call this only in case, when `self.dest.has_capability("hypervisors_heartbeat") is True`. When the candlepin server does not have the capability, then debug message is little bit confusing now.

tests/test_subscriptionmanager.py
```diff
@@ -106,6 +106,21 @@ def test_hypervisorCheckInAsync(self, rhsmconnection):
...
+
```
jirihnidek, line 122: I would add here two more unit tests.
* First unit test for covering the case, when candlepin server does not have the capability.
* Second unit test for the case there are some options: `self.sm.hypervisorHeartbeat(config, options={'global':{'foo': 'bar'}})`

jirihnidek (changes requested):
Overall it is good PR. Thanks. I have few requests.

tests/test_subscriptionmanager.py
wottop: removed call

virtwho/virt/virt.py
wottop: removed debug message

tests/test_subscriptionmanager.py
wottop: this test must be done in Subscription Manager. The connection object is mocked and I cannot test the condition in the connection method from virt-who. Will add subman tests.

jirihnidek: Good point. I just reviewed and merged the PR: https://github.com/candlepin/subscription-manager/pull/2101

jirihnidek (approved):
ACK

PR #219, opened 2019-05-31
1472727: Log error, when encrypted password is missing; ENT-1344

jirihnidek:
* BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1472727
* When key file for encrypted password is missing, then do not
  crash virt-who, but skip given host and print error to log file.
* Added one unit tests for this case.


wottop:
ack

PR #221, opened 2019-06-06
Handle when hyperV instance["BIOSGUID"] returns None causing virt-who to crash

phess:
When the instance's BIOSGUID field is None, a KeyError will not be raised, so None gets passed to decodeWinUUID() which causes an error. Let's treat a BIOSGUID of None like a missing BIOSGUID. Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1718039.

### DIFF
--- a/virtwho/virt/hyperv/hyperv.py
+++ b/virtwho/virt/hyperv/hyperv.py
@@ -583,7 +583,8 @@ class HyperV(virt.Virt):
         for instance in hypervsoap.Pull(uuid):
             try:
                 uuid = instance["BIOSGUID"]
-            except KeyError:
+                assert uuid is not None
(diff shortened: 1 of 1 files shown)


phess:
This is an attempt to fix https://bugzilla.redhat.com/show_bug.cgi?id=1718039 by letting hyperv.py handle scenarios where hyperv returns None instead of raising a KeyError for instance["BIOSGUID"].

wottop:
Please modifiy the commit message and prepend with "1718304:". That way it can be tied to the bug for RHEL 7. 

https://bugzilla.redhat.com/show_bug.cgi?id=1718304

wottop:
ack

PR #222, opened 2019-06-13
Update spec for build system

wottop:


wottop:
The brew system did not like the way the spec file was laid out. This file worked previously, so the brew system must have had a change.

cnsnyder:
@wottop What exactly was the issue in brew? I'm not sure I understand why moving those lines down would change much.

PR #225, opened 2019-06-20
1499679: Check for duplicates in conf file; ENT-249

jirihnidek:
* Warning is printed for Python 3, when there is duplicate key.
  When there are more duplicated keys, then the warning is not
  printed.
* It is not possible to implement this for Python 2 without
  reimplementing parsing of config file.
* Added one unit test for duplicated key
* Fixed one unit test for corrupted encrypted pasword


wottop:
ack

PR #227, opened 2019-06-27
1516120: Handling for incorrect config section headers

wottop:


virtwho/config.py
```diff
@@ -1329,6 +1331,27 @@ def all_drop_dir_config_sections(config_dir=VW_CONF_DIR):
             all_sections.update(parse_file(os.path.join(config_dir, conf)))
         return all_sections
 
+    @staticmethod
+    def has_config_files_in_drop_dir(config_dir=VW_CONF_DIR):
+        """
+        Read all configuration sections in the default config directory
+        :return: a dictionary of {name: {key: value, ... } ... }
+        """
```
jirihnidek, line 1340: Please update the doc string, because current doc string doesn't describe this function.

jirihnidek (changes requested):
Overall it is good. I have only small request.

jirihnidek (approved):
ACK

PR #228, opened 2019-07-11
1720154: Provide SYSTEM_UUID_FACT

pkliczewski:
Hostnames can be changed so we need to make sure to use uniqe identifier.
Kubernetes api provides only machineID which can be used in this context.

Fixes: https://bugzilla.redhat.com/1720154


pkliczewski:
@wottop please review

wottop:
ack

PR #230, opened 2019-08-01
1530254: Update checking on environment variables

wottop:
This check was written before the error handling of configurations was normalized.
It is now redundant and causes entry processing to be short-circuited.

### DIFF
--- a/virtwho/parser.py
+++ b/virtwho/parser.py
@@ -279,20 +279,12 @@ def read_vm_backend_env_variables(env_vars):
 
     if env_vars.get('virt_type') in VM_DISPATCHER.keys():
         virt_type = env_vars['virt_type']
-        try:
-            keys = ['owner', 'server', 'username']
-            for key in keys:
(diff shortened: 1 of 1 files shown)


jirihnidek (changes requested):
Hi Will,
overall it is good solution. I would only add following unit test to `test/test_config_section_virt.py`:

```python
    def test_validate_missing_owner(self):
        """
        Test validation of missing type of virtualization backend
        """
        del self.virt_config['owner']
        result = self.virt_config.validate()
        self.assertIn(('error', 'Required option: "owner" not set.'), result)
        self.assertEqual(len(result), 1)
```

Thoughts?

jirihnidek:
> I would only add following unit test to `test/test_config_section_virt.py`:

Well, this unit test is passing in master too. It has to be somewhere else and something little bit different...

wottop:
That error mesage exists in the previous version also. I am eliminating the error check when the ENV values are put in the config. This test doesn't show a difference.

jirihnidek (approved):
ACK

PR #234, opened 2019-09-16
1733286: add connection/request timeout

pkliczewski:
We need to make sure that kubevirt behaves in the same way as any other plugin. OS level defaults are not enough and we need to respect provided interval. We align the timeout with esxi plugin to 60 seconds even if other plugins use different value.

Bug-Url: https://bugzilla.redhat.com/1733286


wottop:
ack

wottop:
ack

PR #235, opened 2019-09-18
1516209: Proper handling for empty server entry

wottop:


cnsnyder (approved):
Works for me. Thank you!

PR #237, opened 2019-09-20
1751624: Allow unicode username only if python-requests allows it

wottop:
Version 2.20 and above can handle unicode usernames.


jirihnidek (approved):
ACK

PR #240, opened 2019-10-10
1748677: Improved timeout for esx

wottop:
This makes the call to esx a single request each time with no subsequent polling.

Description of maxWaitSeconds option here:
https://code.vmware.com/docs/4205/vmware-vsphere-web-services-sdk-programming-guide/doc/PG_PropertyCollector.7.7.html


chris1984 (approved):
Tested this on my satellite and it works correctly with the same vmware reproducer @wottop was using

jirihnidek (approved):
ACK

PR #246, opened 2020-01-03
1739358 - Cannot read no_proxy value correctly in xen

wottop:


virtwho/virt/xen/xen.py
```diff
@@ -45,6 +45,10 @@ def __init__(self, section_name, wrapper, *args, **kwargs):
         self.add_key('server', validation_method=self._validate_server, required=True)
         self.add_key('username', validation_method=self._validate_username, required=True)
         self.add_key('password', validation_method=self._validate_unencrypted_password, required=True)
+        # strange issue with XEN API and the use of this environment variable
+        no_proxy = self.get('no_proxy','').strip()
```
jirihnidek: There is missing white space. The line should be:
```python
no_proxy = self.get('no_proxy', '').strip()
```

virtwho/virt/xen/xen.py
```diff
@@ -45,6 +45,10 @@ def __init__(self, section_name, wrapper, *args, **kwargs):
         self.add_key('server', validation_method=self._validate_server, required=True)
         self.add_key('username', validation_method=self._validate_username, required=True)
         self.add_key('password', validation_method=self._validate_unencrypted_password, required=True)
+        # strange issue with XEN API and the use of this environment variable
```
jirihnidek, line 48: Interesting. I wonder, how is modified 'no_proxy' used for connecting Xen server? I can see that `server`, `username` and `password` are used, but I cannot see that `no_proxy` is used or anything else from `self.config`.

Is there any documentation or explanation, why should `no_proxy` should start with ','?

I would implement some method for validation `no_proxy` and I would return some info/debug about changed `no_proxy`. Something that we use for in `_validate_server()`. Thoughts?

jirihnidek (changes requested):
Thanks for PR. I request some changes in the code.

virtwho/virt/xen/xen.py
wottop: It appears to be read as an environment variable. When run as a service that means the values in /etc/sysconfig/virt-who.

We do not set it on the connection to Xen via the XenApi. It must just get used quietly. 
The value set in *.conf files is rhsm_no_proxy and only affects the connection for subscription-manager to server communication.

virtwho/virt/xen/xen.py
```diff
@@ -45,6 +45,10 @@ def __init__(self, section_name, wrapper, *args, **kwargs):
...
+        no_proxy = self.get('no_proxy', '').strip()
+        if len(no_proxy) > 0 and not no_proxy.startswith(',') and no_proxy != '*':
+            self.add_key('no_proxy', ',' + no_proxy)
```
jirihnidek, line 51: I don't have any xen server for testing ATM... Is this line really used, because `add_key()` method raises exception, when it is called, without argument `validation_method`. I would like to have at least one unit test for this case.

wottop: [test-xen]
type=xen
server=10.73.5.208
username=root
password=Welcome1!
owner=admin

wottop: closing the PR, does not work

PR #247, opened 2020-01-10
1762780: fix bug, when reporter_id is empty; ENT-1706

jirihnidek:
* BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1762780
* Validation using _validate_non_empty_string did not have side
  effect of setting default value and debug print also didn't
  print default value

### DIFF
--- a/virtwho/config.py
+++ b/virtwho/config.py
@@ -825,9 +825,18 @@ class ConfigSection(collections.MutableMapping):
                 result = ('warning', 'Value for %s not set' % key)
         else:
             if not isinstance(value, str):
-                result = ('warning', '%s is not set to a valid string, using default' % key)
+                if self.has_default(key):
+                    self._values[key] = self.defaults[key]
+                    result = ('warning', '"%s" is not set to a valid string, using default: "%s"' %
+                              (key, self.defaults[key]))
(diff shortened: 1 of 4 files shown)


jirihnidek:
It needs more testing, because I changed validation method.

virtwho/config.py
```diff
@@ -825,9 +825,18 @@ def _validate_non_empty_string(self, key):
                 result = ('warning', 'Value for %s not set' % key)
         else:
             if not isinstance(value, str):
-                result = ('warning', '%s is not set to a valid string, using default' % key)
+                if self.has_default(key):
+                    self._values[key] = self.defaults[key]
+                    result = ('warning', '%s is not set to a valid string, using default: %s' %
+                              (key, self.defaults[key]))
+                else:
+                    result = ('warning', '%s is not set to a valid string and has default value: %s' % key)
```
wottop: has *no* default value?

wottop (changes requested):

virtwho/config.py
jirihnidek: Good point. Thanks.

wottop:
ack

PR #248, opened 2020-01-28
1751441: Add command line path parameter

pkliczewski:
We would like to be able to provide path to kubernetes config file using command line parameter. This PR introduces new `kubevirt-cfg` parameter which can be use to pass the path.


pkliczewski:
@wottop Verified on my centos7

wottop:
ack

PR #249, opened 2020-01-30
1757985: better behavior of virt-who in one shot mode; ENT-1713

jirihnidek:
* BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1757985
* When virt-who is started in one shot mode (option -o) and
  configuration file(s) are provided using -c option(s), then
  at least one configuration file should be valid

### DIFF
--- a/virtwho/main.py
+++ b/virtwho/main.py
@@ -157,6 +157,20 @@ def main():
             logger.error(err)
             exit(1, err)
 
+    if len(effective_config[VW_GLOBAL]['configs']) > 0:
+        # When config file is provided using -c or --config, then other config
+        # files in /etc/virt-who.d are ignored. When it is not possible to read
+        # any configuration file, then virt-who should be terminated
+        cli_config_file_readable = False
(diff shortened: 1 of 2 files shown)


virtwho/main.py
```diff
@@ -157,6 +157,17 @@ def main():
...
             exit(1, err)
 
+    if len(effective_config[VW_GLOBAL]['configs']) > 0:
+        cli_config_file_read = False
+        for file_name in effective_config[VW_GLOBAL]['configs']:
+            if os.path.isfile(file_name):
+                cli_config_file_read = True
+
+        if effective_config[VW_GLOBAL]['oneshot'] is True and cli_config_file_read is False:
+            err = 'No valid configuration file provided in one shot mode'
```
wottop: I don't think that this should be limited to one shot. I know the bug says one shot, but it should apply regardless. If the user specifies a single config file and that file fails, then virt-who should stop.

wottop (changes requested):

wottop:
ack

wottop (approved):

PR #250, opened 2020-02-10
1727203: Validate server name for ASCII only in ESX

wottop:


virtwho/virt/esx/esx.py
```diff
@@ -529,6 +529,14 @@ def __init__(self, *args, **kwargs):
 
     def _validate_server(self, key):
         error = super(EsxConfigSection, self)._validate_server(key)
+        if error is None:
+            try:
```
jirihnidek, line 535: Can you please add here the note mentioned in your comment on bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1727203#c1 It makes 100% sense, but it will not make any sense without this comment after one year ;-).

jirihnidek (changes requested):
Thanks for PR. It look goods, but please add one comment to the code. Thanks.

jirihnidek (approved):
ACK

PR #251, opened 2020-02-11
1780478: Suppress InsecureRequestWarning

wottop:


wottop:
PM determined that the warnings should not be suppressed. Leaving as is.

PR #252, opened 2020-02-12
1780467: Validate rhevm password when unicode is not allowed

wottop:
This will cover all scenarios where a unicode password is not allowed because of the version of the requests package.

### DIFF
--- a/tests/test_config_section_virt.py
+++ b/tests/test_config_section_virt.py
@@ -112,14 +112,24 @@ class TestVirtConfigSection(TestBase):
         result = self.virt_config._validate_unencrypted_password('password')
         self.assertIsNone(result)
 
-    def test_validate_unicode_unencrypted_password(self):
+    @patch('virtwho.config.requests')
+    def test_validate_unicode_unencrypted_password(self, _mock_requests):
         """
         Test of validation of password that is not encrypted and it contains some
         UTF-8 string.
+        Passes or fails based on python-requests version
         """
         self.init_virt_config_section()
         self.virt_config['password'] = 'Příšerně žluťoučký kůň pěl úděsné ódy.'
+        # Version that can't handle unicode
+        _mock_requests.__version__ = '2.19.0'
         result = self.virt_config._validate_unencrypted_password('password')
+        self.assertEqual(result,
+            ('error', "Value: Příšerně žluťoučký kůň pěl úděsné ódy. of option 'password': is not in latin1 encoding."
+                  " That is not allowed on this system."))
+        # Version that can handle unicode
+        _mock_requests.__version__ = '2.21.0'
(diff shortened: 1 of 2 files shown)


tests/test_config_section_virt.py
```diff
@@ -112,14 +112,22 @@ def test_validate_unencrypted_password(self):
...
+    @patch('virtwho.config.requests')
+    def test_validate_unicode_unencrypted_password(self, _mock_requests):
         """
         Test of validation of password that is not encrypted and it contains some
         UTF-8 string.
+        Passes or fails based on python-requests version
         """
         self.init_virt_config_section()
         self.virt_config['password'] = 'Příšerně žluťoučký kůň pěl úděsné ódy.'
+        # Verision that can't handle unicode
```
cnsnyder: Verision -> Version
(both here and below)

tests/test_config_section_virt.py
```diff
@@ -112,14 +112,22 @@ def test_validate_unencrypted_password(self):
...
+        _mock_requests.__version__ = '2.19.0'
         result = self.virt_config._validate_unencrypted_password('password')
+        self.assertIsNotNone(result)
```
cnsnyder: What should the result be if not None?
I would suggest being as specific as we can to ensure proper behaviour.

cnsnyder (changes requested):

cnsnyder:
Also as a general question, how did you decide to check for version 2.20 of requests specifically?

tests/test_config_section_virt.py
```diff
@@ -112,14 +112,24 @@ def test_validate_unencrypted_password(self):
...
+        Passes or fails based on python-requests version
         """
         self.init_virt_config_section()
         self.virt_config['password'] = 'Příšerně žluťoučký kůň pěl úděsné ódy.'
+        # Version that can't handle unicode
+        _mock_requests.__version__ = '2.19.0'
         result = self.virt_config._validate_unencrypted_password('password')
+        self.assertEqual(result,
+            ('error', "Value: Příšerně žluťoučký kůň pěl úděsné ódy. of option 'password': is not in latin1 encoding."
+                  " That is not allowed on this system."))
```
jirihnidek, line 129: @wottop BTW: "Příšerně žluťoučký kůň pěl úděsné ódy" means "The horribly yellowish horse sang terrible odes" in Czech :smiley: and it is widely used for testing encoding in Czech Republic.

jirihnidek (approved):
ACK

cnsnyder (approved):

tests/test_config_section_virt.py
wottop: Thanks for letting me know. I copied it from another test. ;)

PR #255, opened 2020-04-09
1806572: virt-who using V3 APIs for communication with RHEVM which is…

wottop:
… deprecated

The header of Version:3 will still need to be used with a v4 instance.
The url ending of 'ovirt-engine/api' [v4] will be tried before
'api' [older v3] as a fallback.


wottop:
I have the config for a RHEVM4 and a RHEVM3 if you want to smoke test the change.

cnsnyder (approved):
Looks good to me (also smoke tested it and seems to work well).
Thank you for the patch!

PR #257, opened 2020-05-06
1461272: Filter virt-who hosts based on host_parents using wildcard

wottop:


virtwho/virt/esx/esx.py
```diff
@@ -311,6 +307,23 @@ def getHostGuestMapping(self):
             mapping['hypervisors'].append(virt.Hypervisor(hypervisorId=uuid, guestIds=guests, name=name, facts=facts))
         return mapping
 
+    def skip_for_parent(self, host_id, host):
```
jirihnidek, line 310: Please add doc string for this method.

jirihnidek (changes requested):
Looks good. I have only small request. Add doc string to the method.

jirihnidek (approved):
ACK

PR #259, opened 2020-05-14
Update to tests to match changes in Subscription Manager

jirihnidek:


tests/complex/fake_rhevm4.py
```diff
@@ -31,10 +51,19 @@ def check_version_header(self):
...
 
+
 if __name__ == '__main__':
-    rhevm = FakeRhevm4(port=8443)
+    if len(sys.argv) >= 2:
+        port = int(sys.argv[1])
+    else:
+        port = None
+    rhevm = FakeRhevm4(port=port)
+    print(rhevm.virt_who_config())
```
cnsnyder: Why print this? For debug reasons?

tests/complex/fake_sam.py
```diff
@@ -24,15 +50,17 @@ def do_GET(self):
...
     def do_POST(self):
-        print("[FakeSam] POST", self.path)
+        """
+        Handler for POST method
+        """
+        print("[FakeSam] POST", self.path, self.server.code)
         if self.server.code:
             self.write_json({}, status_code=self.server.code, headers={"Retry-After": "60"})
-        elif self.path.startswith('/hypervisors'):
+        elif self.path.startswith('/hypervisors') or self.path.startswith('//hypervisors'):
```
cnsnyder, line 63: Why would it have an extra "/"? Best I could guess is it assumes that there is a hander defined when there isn't one (like it should be set to "candlepin" resulting in "/candlepin/hypervisors".

tests/complex/virtwhotest.py
```diff
@@ -106,7 +141,8 @@ def run_virtwho(self, args, grab_stdout=False, background=False):
         code = None
         data = None
         socket.setdefaulttimeout(1)
-        self.process = Process(target=virtwho.main.main)
+        from virtwho.__main__ import main as main_main
+        self.process = Process(target=main_main)
```
cnsnyder, line 145: Non-nacking comment: This tests slightly less of our codebase as it is a different entry point than that used by the installed virt-who script.

tests/complex/fake_sam.py
wottop: I don't know

tests/complex/fake_rhevm4.py
jirihnidek: It is part of smoke test. When you want to test virt-who against this fake server, then it is good to see the content of configuration file that could be used for testing.

PR #265, opened 2020-08-25
1872477: virt-who fails to parse output from hypervisor

RudneiBertolJr:
Added a decode to the function that format the message to string.


wottop:
This PR has been replaced by https://github.com/candlepin/virt-who/pull/275
PR 275 decodes the data in the field that is causing the problem and allows that cleaned data to be used in the facts.

PR #266, opened 2020-09-09
184506: virt-who should send its version in the User-Agent header

wottop:
My attempts to create a unit test were not successful. I stopped in the interest of time.

The way to test this is to run candlepin with debug logging and see the user-agent header there.


jirihnidek (approved):

jirihnidek:
I smoke tested this using environment variables `$SUBMAN_DEBUG_PRINT_REQUEST` and `$SUBMAN_DEBUG_PRINT_REQUEST_HEADER`

PR #267, opened 2020-09-22
1874371: Correctly gather hostname for libvirt instances

wottop:
Update print option to display data format that is sent to server


wottop:
Moved to other PR's

PR #268, opened 2020-09-23
1878136: Deprecation warning for environment variables

wottop:
Will show either when known environment variables are used, or
 or when the sysconfig file is present

### DIFF
--- a/virt-who.8
+++ b/virt-who.8
@@ -25,7 +25,7 @@ Use configuration file directly (will override configuration from other files. '
 .IP
 
 .SH ENVIRONMENT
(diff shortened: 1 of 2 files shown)


jirihnidek (changes requested):
Hi, I smoke tested and it works as expected, but there is some problem with unit tests. One test is failing on Travis and I can confirm that the py.test is for some reason, when I run: `py.test tests/test_subscriptionmanager.py::TestSubscriptionManagerConfig::test_sm_config_env`. Maybe it will be necessary to mock logging in tests/test_subscriptionmanager.py::TestSubscriptionManagerConfig.

jirihnidek (approved):
ACK

PR #269, opened 2020-09-29
Add patch to remove AHV bits for RHEL builds

wottop:
Scratch builds:

Fedora: https://koji.fedoraproject.org/koji/taskinfo?taskID=52472283
RHEL: https://brewweb.engineering.redhat.com/brew/taskinfo?taskID=31613797


jirihnidek (approved):
I can confirm that rpm is built with support of AHV on Fedora 32 and the RPM is built without support of AHV on RHEL8.

PR #275, opened 2020-10-19
1876927: virt-who fails to parse output from hypervisor

wottop:
The cluster name in vCenter allows characters that require
 URL encoding. Because of this, that value needs to be URL
 decoded before the data can be used.

### DIFF
--- a/tests/test_esx.py
+++ b/tests/test_esx.py
@@ -314,6 +314,64 @@ class TestEsx(TestBase):
         result = self.esx.getHostGuestMapping()['hypervisors'][0]
         self.assertEqual(expected_result.toDict(), result.toDict())
 
+
+    @patch('suds.client.Client')
+    def test_getHostGuestMapping_cluster_slash(self, mock_client):
+        expected_hostname = 'hostname.domainname'
+        expected_hypervisorId = 'Fake_uuid'
+        expected_guestId = 'guest1UUID'
+        expected_guest_state = Guest.STATE_RUNNING
+
+        fake_parent = MagicMock()
+        fake_parent.value = 'fake_parent_id'
+        fake_parent._type = 'ClusterComputeResource'
+
+        fake_vm_id = MagicMock()
+        fake_vm_id.value = 'guest1'
+
+        fake_vm = MagicMock()
+        fake_vm.ManagedObjectReference = [fake_vm_id]
+        fake_vms = {'guest1': {'runtime.powerState': 'poweredOn',
+                               'config.uuid': expected_guestId}}
+        self.esx.vms = fake_vms
+
+        fake_host = {'hardware.systemInfo.uuid': expected_hypervisorId,
+                     'config.network.dnsConfig.hostName': 'hostname',
+                     'config.network.dnsConfig.domainName': 'domainname',
+                     'config.product.version': '1.2.3',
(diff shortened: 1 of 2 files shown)


virtwho/virt/esx/esx.py
```diff
@@ -40,6 +40,11 @@
 from virtwho import virt
 from virtwho.config import VirtConfigSection
 
+try:
+    from urllib import unquote as urldecode
+except:
```
jirihnidek: Please do not use generic exception, but use: `except ImportErrror:`

tests/test_esx.py
```diff
@@ -314,6 +314,64 @@ def test_getHostGuestMapping_incomplete_data(self, mock_client):
...
+                     'config.product.version': '1.2.3',
+                     'hardware.cpuInfo.numCpuPackages': '1',
+                     'name': expected_hostname,
+                     'parent': fake_parent,
+                     'vm': fake_vm,
+                     }
+        fake_hosts = {'random-host-id': fake_host}
+        self.esx.hosts = fake_hosts
+
+        fake_cluster = {'name': 'SAG-AFTRA Dev%2fTest Cluster'}
```
jirihnidek: I would change it to something to:
```python
fake_cluster = {'name': 'FOO-BAR Dev%2fTest Cluster'}
```

tests/test_esx.py
```diff
@@ -314,6 +314,64 @@ def test_getHostGuestMapping_incomplete_data(self, mock_client):
...
+                    expected_guestId,
+                    self.esx.CONFIG_TYPE,
+                    expected_guest_state,
+                )
+            ],
+            facts={
+                Hypervisor.CPU_SOCKET_FACT: '1',
+                Hypervisor.HYPERVISOR_TYPE_FACT: 'vmware',
+                Hypervisor.HYPERVISOR_VERSION_FACT: '1.2.3',
+                Hypervisor.HYPERVISOR_CLUSTER: 'SAG-AFTRA Dev/Test Cluster',
```
jirihnidek: And I would use here:
```python
Hypervisor.HYPERVISOR_CLUSTER: 'FOO-BAR Dev/Test Cluster',
```

virtwho/virt/esx/esx.py
```diff
@@ -40,6 +40,11 @@
 from virtwho import virt
 from virtwho.config import VirtConfigSection
 
+try:
+    from urllib import unquote as urldecode
```
jirihnidek: I guess that this import is intended for Python2.7. Do we plan to still support Python2.7?
Anyway, shouldn't we try to import module for Python3 and if it is not possible, then try to import module for Python2.7?

jirihnidek (changes requested):
Overall it is good approach. The `unquote()` method seems as stable and fail tolerant. I have some small requests.

virtwho/virt/esx/esx.py
wottop: We will likely backport this to 7.9 at some point so I think it is worth leaving. I did switch the order of the imports.

jirihnidek: Oh, in this case it make sense.

jirihnidek (approved):
ACK

PR #277, opened 2020-10-21
1658440: New section of virt-who.conf file for environment variables

wottop:


wottop:
Closing because there is a separate card for the 8.4 version. In 8.4 this is added, but the sysconfig/virt-who file will remain.

PR #278, opened 2020-10-22
1890421: New section of virt-who.conf file for environment variables

wottop:


virtwho/config.py
```diff
@@ -1451,6 +1452,19 @@ def _check_effective_config_validity(effective_config, config_files_not_complete
     effective_config.validate()
     return effective_config, validation_errors
 
+def parse_system_environment(config):
+    """
+    :param config: The dict of parameters to be added to the system environment
+    """
+    has = False
+    keys = []
+    for key,value in config.items():
```
jirihnidek: Please add space between `key,` and `value`

virtwho/config.py
```diff
@@ -1451,6 +1452,19 @@ def _check_effective_config_validity(effective_config, config_files_not_complete
...
+        has = True
+        os.environ[key]=value
+        keys.append(key)
+    if has:
+        logger.debug("Environment variables added from system_environment section of " +
```
jirihnidek: I would change the message a little bit to highlight system_environment section:

```python
"Environment variables added from [system_environment] section of "
```

virtwho/config.py
```diff
@@ -1451,6 +1452,19 @@ def _check_effective_config_validity(effective_config, config_files_not_complete
...
+                     "general configuration file: %s" % keys)
```
jirihnidek: When only one key is present in system_environment section, then it prints something like this:

`Environment variables added from system_environment section of general configuration file: ['https_proxy']`

So, I would change this line little bit too:

```python
"general configuration file: %s" % ', '.join(keys)
```

template-general.conf
```diff
@@ -32,6 +32,17 @@
 #owner=                 ; owner for use with SAM, Customer Portal, or Satellite 6
 #hypervisor_id=         ; how will be the hypervisor identified, one of: uuid, hostname, hwuuid
 #rhsm_no_proxy=         ; should not use a proxy for specific hosts.
+#
+#[system_environment]   ; Values in this section will be copied to the system environment and will be in place
+#                       ; for the extent of the execution.
+#                       ; There are a few suggested options, but any name/value intended for the system environment
+#                       ; can be put here. Other than the copy process, virt-who will not read or specifically react
+#                       ; to them.
+#https_proxy            ; Proxy hostname for all https requests
```
jirihnidek: What about some examples? You can use: 'https://foo.bar.com:3128' or 'foo.bar.com'

tests/test_config.py
```diff
@@ -1040,3 +1040,19 @@ def test_special(self):
...
             ['abc ', '\\ def', 'g h i', ' jkl ']
         )
+
+# Values used for testing system_environment
+SYSTEM_ENVIRONMENT_SECTION_VALUES = {
+    'http_proxy': 'this.proxy.com',
+    'https_proxy': 'that.proxy.com',
+    'no_proxy': 'not.this.proxy.com'
+}
+
```
jirihnidek, line 1051: Please add here one more empty line.

virtwho/config.py
```diff
@@ -1451,6 +1452,19 @@ def _check_effective_config_validity(effective_config, config_files_not_complete
     effective_config.validate()
     return effective_config, validation_errors
 
+def parse_system_environment(config):
```
jirihnidek, line 1456: Please add one empty line before and after this function.

tests/test_config.py
```diff
@@ -1040,3 +1040,19 @@ def test_special(self):
             parse_list("'abc\ ', '\\\\ def' ,'g h i' , ' jkl '"),
             ['abc ', '\\ def', 'g h i', ' jkl ']
         )
+
```
jirihnidek, line 1043: Please add one more empty line here to fix stylish issue.

jirihnidek (commented):
@wottop Thanks for the PR. It works for most cases. I found one case, when it doesn't work as I would expect. I also added some comments to the code.

1. Set environment variable: `export HTTP_PROXY=foo.bar.com`
2. Set `no_proxy` to `foo.bar.com` in configuration file `/etc/virt-who.conf`
3. Try to start virt-who

Expected result:
Host-to-guest mapping is gathered and reported to candlepin server

Actual result:
The virt-who tries to use not-available proxy server `foo.bar.com`

Note: When `no_proxy` is set to `*`, then it works as expected.

jirihnidek (approved):
ACK

PR #280, opened 2020-11-09
Do not use deprecated isAlive() but use is_alive()

jirihnidek:
* This becomes real issue on Fedora 33 and Python 3.9, because
  isAlive() is not here supported anymore.


wottop:
ack

cnsnyder (approved):
Looks good.  Thank you!
(TIL the "is_alive" spelling has been around since python 2.6)

PR #289, opened 2020-12-18
1855550: [Remote Libvirt] The Name in Stage Candlepin cannot update b…

cnsnyder:
…ased on hypervisor_id configuration

The hostname was not collected correctly in the remote libvirt scenario for the name field.
Katello uses the hypervisor id on creation as a fallback for this field which is
then stored in Katello. It does not use this fallback method on updates, so a change in
hypervisor_id does not show in Katello. This fix remedies the original
problem where the hostname was not assigned to the name field in the hypervisor report.


wottop:
ack

PR #290, opened 2020-12-18
1909145: [RFE] Use single json format for input/output data

wottop:
Update print option to display data format that is sent to server
Update fakevirt to receive data format that is sent to server


jirihnidek (approved):
ACK

PR #291, opened 2021-01-05
1910020: replace deprecated call

wottop:
Element.getchildren was removed entirely in Python 3.9

### DIFF
--- a/virtwho/virt/hyperv/hyperv.py
+++ b/virtwho/virt/hyperv/hyperv.py
@@ -404,12 +404,11 @@ class HyperVSoap(object):
     def _Instance(cls, xml_doc):
         def stripNamespace(tag):
             return tag[tag.find("}") + 1:]
-        children = xml_doc.getchildren()
-        if len(children) < 1:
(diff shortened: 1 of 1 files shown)


virtwho/virt/hyperv/hyperv.py
```diff
@@ -404,12 +404,12 @@ def post(self, body):
     def _Instance(cls, xml_doc):
         def stripNamespace(tag):
             return tag[tag.find("}") + 1:]
-        children = xml_doc.getchildren()
+        children = xml_doc
```
jirihnidek: I think that it is useless to have another variable `children`. I would just use `xml_doc` in all cases, where `children` is used in this method.

jirihnidek (changes requested):
Hi, thanks for PR. I have only small request.

BTW: I checked that `getchildren()` or any other deprecated method from `xml` package is not used in `virt-who` nor `sub-man`.

jirihnidek (approved):
ACK

PR #292, opened 2021-01-14
handle not running vms

pkliczewski:
There are several phases that vms can be in. Only in Running phase there is hostName available. This patch improves how we handle other phases.

There are potential issues with cert verification when using self signed ca. In this PR we add new param to connect without cert verification and host validation.

### DIFF
--- a/virt-who-config.5
+++ b/virt-who-config.5
@@ -145,6 +145,9 @@ Path to Kubernetes configuration file which contains authentication and connecti
 .TP
 \fB#kubeversion\fR
 API version used to override kubevirt api version fetched from the cluster. Used by kubevirt option
+.TP
+\fB#insecure\fR
+Eliminate validation of tls certificates during connection to kubevirt
 
 .SH EXAMPLE
 [test-esx]
--- a/virtwho/virt/kubevirt/kubevirt.py
+++ b/virtwho/virt/kubevirt/kubevirt.py
@@ -24,7 +24,7 @@ import os
 import os.path
 
 from virtwho import virt
-from virtwho.config import VirtConfigSection
+from virtwho.config import VirtConfigSection, str_to_bool
 from virtwho.virt.kubevirt.client import KubeClient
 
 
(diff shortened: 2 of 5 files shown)


pkliczewski:
@wottop  @dankenigsberg please take a look

virtwho/virt/kubevirt/kubevirt.py
```diff
@@ -136,12 +141,12 @@ def getHostGuestMapping(self):
 
         for vm in vms['items']:
             spec = vm['spec']
-            host_name = vm['status']['nodeName']
 
             # a vm is not scheduled on any hosts
-            if host_name is None:
+            if vm['status']['phase'] != 'Running':
```
dankenigsberg: Frankly, virt-who does not really care if the VM is `Running` it only cares if it is mapped to a host. Thus I think that 
```
host_name = vm['status'].get('nodeName')
```
would be a more precise change.

dankenigsberg (changes requested):
Thanks for fixing these and adding the tests!

I'd have two different commits for the two different issues, but it's not my project to enforce.

virtwho/virt/kubevirt/kubevirt.py
pkliczewski: As far as I understand how k8s work when a vm is `Running` phase `nodeName` always will be there. Do you suggest to make the changes as minimal as possible?

dankenigsberg: No, my issue is of precision. If for some reason KubeVirt has a bug and does not report the node on a Running VM. "My" suggested code just skips it. If KubeVirt introduces a new status "RunningFast", I don't care. virt-who needs to report mapping of VMs to nodes; it should not care about anything else.

pkliczewski: OK, I see your point. Will update this PR. Thanks!

pkliczewski:
@wottop What do you think?

virt-who-config.5
```diff
@@ -145,6 +145,9 @@ Path to Kubernetes configuration file which contains authentication and connecti
 .TP
 \fB#kubeversion\fR
 API version used to override kubevirt api version fetched from the cluster. Used by kubevirt option
+.TP
+\fB#insecure\fR
+Allows not to validate tls certificates during connection to kubevirt
```
wottop: Eliminate validation of tls certificates ...

wottop (changes requested):
Just one suggested change in the man page

virt-who-config.5
pkliczewski: Done

wottop:
ack

PR #295, opened 2021-02-01
1922945: Fixing a possible typo in applyClusterComputeResource

amasolov:
I do believe that there is a typo in the line I want to fix. Have been having hard time to find any possible documentation about these events/kinds. I would appreciate any links to them. However, looking at other similar calls like: applyVirtualMachineUpdate and applyHostSystemUpdate, I assume that 'kind' was a typo and it should be 'modify'. Definitely need someone who knows what's going on to have a look into this.


wottop:
As this has not been seen by other users or the people testing virt-who, I am going to close this PR.
If you can create a reproducer for the problem. please reopen the accompanying BZ so we can work on it.

PR #296, opened 2021-02-16
Man page update to describe the migration script [ENT-3401]

wottop:


virt-who-config.5
```diff
@@ -249,6 +249,27 @@ Absolute path to the JSON file that has the same structure as file returned from
 .br
 }
 
+.SH CONFIGURATION MIGRATION
+
+Previous versions of virt-who employed other configuration means based on environment variables. The first was system-wide environment variables. The second was a file [/etc/sysconfig/virt-who]
+which contained variables to be used on a session basis when run as a service.
+
+In an effort to simplify the process, and to bring consistency to virt-who's execution regardless of initiation [service vs. command line], the configuration of the application has been
+consolidated into the general configuration file as outlined above. To update previous versions of the configuration layout, a migration script has beeen included. That script will incorporate
```
cnsnyder: A few suggestions below:
1) beeen -> been.
2) 
```The configuration of virt-who has been streamlined. Previously it was possible to configure virt-who using 4 methods (potentially combined, in this following order of preference): command-line arguments & options, environment variables and virt-who configuration files (/etc/virt-who.conf. /etc/virt-who.d/*.conf). Environment variables were taken into account regardless of how they were set, directly in the execution of the virt-who daemon, or loaded into the environment via /etc/sysconfig/virt-who.

The new version of virt-who no longer supports setting most options via the environment. In order to not loose previously valid configurations that made use of environment variables specified in /etc/sysconfig/virt-who, a migration script has been added...```

File: virt-who-config.5 [hunk 2]
```diff
@@ -249,6 +249,27 @@ Absolute path to the JSON file that has the same structure as file returned from
...
+the known system environment variables [VIRTWHO_INTERVAL, VIRTWHO_DEBUG, VIRTWHO_ONESHOT] and the entries in the /etc/sysconfig/virt-who file into the general configuration file. The known variables
+will land in the \fBglobal\fR section while any others in the /etc/sysconfig/virt-who [i.e. HTTPS_PROXY] will land in the \fBsystem_environment\fR section. This migration may result
+in multiple entries for a specific fields, but the priority order will be maintained so that the behavior will remain constant.
+
+Each new entry in the general configuration file will come after a comment indicating that it was migrated. The /etc/sysconfig/virt-who file will be deleted after its entries are
+migrated, but the known system environment variables  will need to be manually removed or they will be migrated again if the script is rerun. Those variables will not be recognized
+by virt-who even if they remain.
+
+The spec file will cause the migration script to run when the RPM is installed, but you can run it manually with python after the RPM is installed:
```
cnsnyder: The migration script is run on upgrades to newer versions of virt-who.

virt-who-config.5
```diff
@@ -249,6 +249,27 @@ Absolute path to the JSON file that has the same structure as file returned from
...
+consolidated into the general configuration file as outlined above. To update previous versions of the configuration layout, a migration script has beeen included. That script will incorporate
+the known system environment variables [VIRTWHO_INTERVAL, VIRTWHO_DEBUG, VIRTWHO_ONESHOT] and the entries in the /etc/sysconfig/virt-who file into the general configuration file. The known variables
+will land in the \fBglobal\fR section while any others in the /etc/sysconfig/virt-who [i.e. HTTPS_PROXY] will land in the \fBsystem_environment\fR section. This migration may result
+in multiple entries for a specific fields, but the priority order will be maintained so that the behavior will remain constant.
+
+Each new entry in the general configuration file will come after a comment indicating that it was migrated. The /etc/sysconfig/virt-who file will be deleted after its entries are
+migrated, but the known system environment variables  will need to be manually removed or they will be migrated again if the script is rerun. Those variables will not be recognized
```
jirihnidek: This part could be little bit confusing. It describes the case, when environment variables are not specified in /etc/sysconfig/virt-who. I would emphasize that fact. Maybe it would be good to write some examples how environment variables could be specified in other cases (/etc/profile, /etc/profile.d/*, etc.)

virt-who-config.5
```diff
@@ -249,6 +249,27 @@ Absolute path to the JSON file that has the same structure as file returned from
 .br
 }
 
+.SH CONFIGURATION MIGRATION
+
+Previous versions of virt-who employed other configuration means based on environment variables. The first was system-wide environment variables. The second was a file [/etc/sysconfig/virt-who]
```
jirihnidek: I guess that the 'means' word was added by mistake in the first sentence, right?

virt-who-config.5
jirihnidek: I would not mention spec file here, because administrator of systems needn't to have experience with creation of RPM and referencing .spec file could be confusing. I would write here something like: 'The migration script will be run, when the RPM is installed, but ...'

jirihnidek (changes requested):
I looked at man page using man CLI. It looks good. I have few suggestions to text.

virt-who-config.5
```diff
@@ -249,6 +249,27 @@ Absolute path to the JSON file that has the same structure as file returned from
...
+which contained variables to be used on a session basis when run as a service.
+
+In an effort to simplify the process, and to bring consistency to virt-who's execution regardless of initiation [service vs. command line], the configuration of the application has been
+consolidated into the general configuration file as outlined above. To update previous versions of the configuration layout, a migration script has beeen included. That script will incorporate
+the known system environment variables [VIRTWHO_INTERVAL, VIRTWHO_DEBUG, VIRTWHO_ONESHOT] and the entries in the /etc/sysconfig/virt-who file into the general configuration file. The known variables
+will land in the \fBglobal\fR section while any others in the /etc/sysconfig/virt-who [i.e. HTTPS_PROXY] will land in the \fBsystem_environment\fR section. This migration may result
+in multiple entries for a specific fields, but the priority order will be maintained so that the behavior will remain constant.
```
cnsnyder: i'd suggest "...multiple entries for a given field..." (less the "...")

virt-who-config.5
```diff
@@ -249,6 +249,27 @@ Absolute path to the JSON file that has the same structure as file returned from
...
+by virt-who even if they remain.
+
+The spec file will cause the migration script to run when the RPM is installed, but you can run it manually with python after the RPM is installed:
+
+[python_sitelib]/virtwho/migrate/migrateconfiguration.py
+
+
+
 .SH AUTHOR
 Radek Novacek <rnovacek at redhat dot com>
```
cnsnyder, line 275: Should this be updated to include Will?

cnsnyder (changes requested):

virt-who-config.5
```diff
@@ -249,8 +249,32 @@ Absolute path to the JSON file that has the same structure as file returned from
...
+Setting of environment variables [set in the user's profile script or in the default global profile]
+.br
+A service environment file [/etc/sysconfig/virt-who].
+
+The new version of virt-who no longer supports setting most options via the environment. In order to not lose previously valid configurations that made use of environment variables,
+a migration script has been added. That script will incorporate the known system environment variables [VIRTWHO_INTERVAL, VIRTWHO_DEBUG, VIRTWHO_ONESHOT] and the entries in the
+service environment file into the general configuration file. The known variables will land in the \fBglobal\fR section while any others in the service environment file [i.e. HTTPS_PROXY]
+will land in the \fBsystem_environment\fR section. This migration may result in multiple entries for a specific field.
+
+Each new entry in the general configuration file will come after a comment indicating that it was migrated. The service environment file file will be deleted after its entries are
```
jirihnidek: Two words "file" in the sentence. I guess that there should be: "The service environment file /etc/sysconfig/virt-who will be deleted after its entries are..."

virt-who-config.5
jirihnidek: I agree :+1:

jirihnidek (changes requested):
One small change is necessary.

virt-who-config.5
wottop: I used the wording "service environment file [/etc/sysconfig/virt-who]" the first time in hopes that I would not need to specify the path in subsequent references.

cnsnyder (approved):

PR #298, opened 2021-02-18
1924572: Add insecure option to config template

wottop:


cnsnyder (approved):
Looks good to me! Thank you!

PR #300, opened 2021-05-11
1920322: Uncomment section header on migrate

wottop:


jirihnidek:
@wottop Hi, the bug number is not correct, because 1920332 refers to `Fedora` and `fedora-repos` component: https://bugzilla.redhat.com/show_bug.cgi?id=1920332

virtwho/migrate/migrateconfiguration.py
```diff
@@ -74,13 +74,20 @@ def migrate_env_to_config(sysconfig_filename=None, general_config_filename=None)
     has_global = False
 
     for line in lines:
-        if line.startswith('[global]'):
+        stripped = line.strip()
+        if stripped == '[global]' or stripped == '#[global]':
```
jirihnidek: What about these cases: `# [global]` or `#  [global]` etc.?

jirihnidek (changes requested):
Looks good. I have only small concerns about about various variants of commenting sections. Not sure how serious it is.

jirihnidek (approved):
Thanks for update. This will be more reliable. :+1:

PR #303, opened 2021-05-17
Remove Xen from hypervisor types

wottop:


virtwho/virt/libvirtd/libvirtd.py
```diff
@@ -375,7 +375,7 @@ def _listDomains(self):
                     continue
 
                 if domain.UUIDString() == "00000000-0000-0000-0000-000000000000":
-                    # Don't send Domain-0 on xen (zeroed uuid)
+                    # Don't send Domain-0 (zeroed uuid)
```
jirihnidek, line 378: Could be this zeroed uuid set in any case and should we ignore it? I don't understand why is this code (related to xen) in `libvirtd.py`.

jirihnidek (changes requested):
Thanks for PR. I can confirm that `xen` is not supported, but when I tried to use "type=xen" in my configuration file, then I didn't see any relevant error/warning message in log file that I try to use unsupported virtualization backend. I think that we should print some warning message that this virtualization backend is not supported anymore.

I was also able to find "xen" keywords in following files:
* build-rpm-no-ahv.patch
* tests/test_virtwho.py
* virt-who-config.5
* virt-who.8

And I have one more question/comment in the code.

virtwho/virt/libvirtd/libvirtd.py
wottop: I do not know why this is in there as this code is quite old. I left in the condition so that it would still behave the same way. It may never occur now that Xen is gone, but I wanted to be safe.

wottop:
build-rpm-no-ahv.patch - this will get cleaned up when the patch is remade after this branch is merged.
tests/test_virtwho.py - the text is 'xenv' which stands for 'x env'. it is not a reference to xen.

wottop:
Note that there is already a test for unknown 'type' values. It still passes with my change to incorporate an error in the logging for an unknown type value.

virtwho/virt/libvirtd/libvirtd.py
jirihnidek: OK, probability that random UUID will have this value is really really low. :-)

jirihnidek:
> Note that there is already a test for unknown 'type' values. It still passes with my change to incorporate an error in the logging for an unknown type value.

Yeah, I can see the test (file `test_config.py` and method `test_invalid_type()`). I can also see following error message in log now:

```
2021-05-20 09:23:19,324 [virtwho.rhsm_log ERROR] MainProcess(205135):MainThread @config.py:init_config:1601 - [xen_experiment]: Unsupported virtual type 'xen' is set
```
:+1:

jirihnidek (approved):
ACK. RIP Xen support.

PR #309, opened 2021-05-26
1943486: Clear report cache when no hypervisors are reported

wottop:


wottop:
To test you will need to follow the details on the card for adding/removing hypervisors from the vCenter instance:

https://issues.redhat.com/browse/ENT-3703

PR #310, opened 2021-05-27
1965320: Clear previous report hash when hypervisor count is zero

wottop:


virtwho/virt/virt.py
```diff
@@ -710,6 +710,13 @@ def _send_data(self, data_to_send):
                 for source_key in reports_batched:
                     self.submitted_report_and_hash_for_source[source_key] =\
                         (batch_host_guest_report, data_to_send[source_key].hash)
+        else:
```
jirihnidek: What about the case, when virt-who is configured to gather host-to-guest mapping from more hypervisors? When other hypervisor will report some mapping, then this bug fix will not work for this case. The cache will not be cleared. I don't know what is included in `data_to_send`, when one hypervisor do not have any quest. Is there any instance of `ErrorReport` or anything else or this hypervisor without any guest is not included there at all?

jirihnidek (changes requested):
Thanks for the PR. I suspect that there could be problem, when virt-who is configured to gather mapping from more hypervisors. I don't have system for testing this case. Please look at my comment.

virtwho/virt/virt.py
wottop: When it batches, it only includes the reports that show a change. I tested it with 2 configs, each with one hypervisor. Config 1's hypervisors never change, so it only gets included in the batched report the first time. I remove the hypervisor from config 2 and the cache gets cleared because it is in the batch by itself. When I put the hypervisor back in config 2 then it reports on that hypervisor correctly [which is where the bug was]. 

The issue you describe is now only possible [with this fix] if at least one of the other configs has a change for every cycle and then does not allow the zero hypervisors scenario in the batch. Note that the bug only surfaces if you remove a hypervisor [the last one] for a config and then put back the same hypervisor with the same data [hosts and guests].  In actuality, the lack of a report in the virt-who log for the readded hypervisor is where the complaint lies. Candlepin still has the hypervisor in the DB because this is all additive anyway. We never remove a hypervisor because of an absence in the report. It does not lead to bad data, it just doesn't 'feel right'. If the readded hypervisor had different guests, for instance, then the report goes out regardless.

Let me know if I can explain this better.

wottop: Updated to handle on a single report instead of all in batch.

jirihnidek (approved):
ACK

PR #312, opened 2021-06-04
Update certs for complex tests

wottop:
Was getting an error for the encryption level in the pytests.


wottop:
retest this please

jirihnidek (approved):
Thanks for the PR. I can see that certificate uses SHA-256, which is much better thank SHA-1. I can see two more unit tests failing on Fedora34, but it is not related to cert/key. We can address two other unit tests in another PR.

PR #314, opened 2021-06-14
Record last dates of succcess for sources and destinations

wottop:


jirihnidek:
Some unit tests are failing on Jenkins. Complex tests are failing due to timeout. I tried to restart job, but it ended with the same result. When I run unit tests locally, then some tests are failing too, but for little bit different reason:

http://pastebin.test.redhat.com/971745

I get these errors for main branch too.

It seems that the certificate used for complex test `./tests/complex/cert.pem` uses too weak hash function (SHA1 in this case). We should regenerate this certificate to use something more robust. I propose to use some modern hash function, e.g. SHA-512.

virtwho/pid_lock.py
```diff
@@ -0,0 +1,65 @@
...
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""
+
```
jirihnidek: We need stylish check for virt-who too on the Jenkins. This file contains too many stylish issues.

I know that you copy pasted most of the code from some other place, but please correct style of the code.

virtwho/executor.py
```diff
@@ -68,8 +74,36 @@ def _create_virt_backends(self):
...
             virts.append(virt)
+        self._init_run_status(config_names)
         return virts
 
+    def _init_run_status(self, config_names=[]):
+        # ensure that there is an entry in the status file for each config. We don't care if they work, we
+        # need to record their existence
+        lock = PIDLock(STATUS_LOCK)
+        while lock.is_locked():
+            time.sleep(1)
```
jirihnidek: I think that this waiting while loop should not be here. When lock is locked, then following `with lock` is blocked until `lock` is unblocked.

virtwho/executor.py
```diff
@@ -68,8 +74,36 @@ def _create_virt_backends(self):
...
+        with lock:
+            os.makedirs(STATUS_DATA_DIR, exist_ok=True)
+            # need to create the file if it does not exist
+            with open(STATUS_DATA, "w+") as json_status:
```
jirihnidek: We read the JSON file here. Thus the file should be opened using in read ("r") mode not "w+".

virtwho/executor.py
```diff
@@ -68,8 +74,36 @@ def _create_virt_backends(self):
...
+                    status_dict = json.load(json_status)
+                except JSONDecodeError:
+                    status_dict = {}
+            if 'sources' not in status_dict:
+                status_dict['sources'] = {}
+            if 'destinations' not in status_dict:
+                status_dict['destinations'] = {}
+            for name in config_names:
+                if name not in status_dict['sources']:
+                    status_dict['sources'][name] = {"last_successful_retrieve": "none"}
```
jirihnidek: I believe that the value should be `None` and not "none", because we want to store `null` value to JSON file, right? Or do we want to store "none" string there?

virtwho/executor.py
```diff
@@ -68,8 +74,36 @@ def _create_virt_backends(self):
...
+                if name not in status_dict['destinations']:
+                    status_dict['destinations'][name] = {"last_successful_send": "none","last_job_id": "none"}
```
jirihnidek: Same comment here and there is missing space between comma and "last_job_id".

virtwho/pid_lock.py
```diff
@@ -0,0 +1,65 @@
...
+import sys
+import os
+
+PIDFILE = "/var/run/virt-who.pid"
+STATUS_LOCK="/var/run/virt-who-status.pid"
```
jirihnidek: Missing empty space around `=`

virtwho/pid_lock.py
```diff
@@ -0,0 +1,65 @@
...
+STATUS_DATA="/var/lib/virt-who/run_data.json"
+STATUS_DATA_DIR="/var/lib/virt-who"
+
```
jirihnidek: Please add empty line here.

virtwho/virt/virt.py
```diff
@@ -807,6 +826,17 @@ def check_report_status(self, report):
             # If we get here and have to try again, it's not our first rodeo...
             first_attempt = False
 
+    def record_status(self, config_name, type, json_info):
+        lock = PIDLock(STATUS_LOCK)
+        while lock.is_locked():
+            time.sleep(1)
```
jirihnidek: I believe that this while loop needn't to be here too.

virtwho/virt/virt.py
```diff
@@ -807,6 +826,17 @@ def check_report_status(self, report):
             # If we get here and have to try again, it's not our first rodeo...
             first_attempt = False
 
+    def record_status(self, config_name, type, json_info):
```
jirihnidek, line 829: Please add doc string to this method.

jirihnidek (changes requested):
Thanks for the PR. I can confirm that it works as expected for simple case. I have few requests.

virtwho/executor.py
wottop: Turns out that the original use of the PIDLock was only to stop from running more than one virt-who process. It did not wait to get the lock, but just reported it and stopped the process. 
I put the wait loop into the PIDLock class and made it conditional so as to not disrupt the original execution. When the wait loop was completely removed, then two processes would enter the 'with lock' block at the same time.

virtwho/executor.py
wottop: w+ plus is used so that the file is created when it does not exist.
I added to the comment.

jirihnidek: OK. Make sense.

virtwho/pid_lock.py
```diff
@@ -0,0 +1,72 @@
...
+                os.remove(self.filename)
+                return False
+        except Exception:
+            return False
+
+    def __enter__(self):
+        # ensure file not locked
+        if self.wait_to_lock:
+            while self.is_locked():
+                time.sleep(1)
```
jirihnidek: It seems that `PIDLock` was implemented to avoid running multiple instances of virt-who. 

I think that this approach has danger of race conditions, because there is no atomicity of check (file exists) and set (save current PID to file). There is no timeout in while loop and one process can be stuck forever. I'm not sure about sleeping for one second.

I would rather use approach described here:

https://stackoverflow.com/questions/489861/locking-a-file-in-python

Example:

```python
from filelock import FileLock
import os
import time


def main():
    pid = os.getpid()
    print(f'{pid} is waiting for lock')
    with FileLock("myfile.txt.lock"):
        print(f"{pid} lock acquired...")
        with open("myfile.txt", "w+") as my_file:
            my_file.write(str(pid))
            time.sleep(2.0)
    print(f"{pid} lock released")


if __name__ == '__main__':
    main()
```

and then try to run it using: `python ./example.py & python ./example.py`

jirihnidek: Oh, this is not part of Python. It requires another RPM :-(.

jirihnidek (commented):
I have concerns about locking of `/var/run/virt-who-status.pid`. Because virt-who runs only on Linux, than we can use `fcntl.flock` for locking the file.

It would be nice to be able to use `filelock` package, but I'm not sure if we can add another dependency on the `filelock`. If not, then it could be at least an inspiration for us. Thoughts?

virtwho/lock.py
```diff
@@ -0,0 +1,108 @@
...
+        except Exception as e:
+            print("Unable to create pid file: %s" % str(e), file=sys.stderr)
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        try:
+            os.remove(self.filename)
+        except Exception:
+            pass
+
+class FileLock(object):
```
jirihnidek, line 73: I would implement `__enter__` and `__exit__` method and I would make make `acquire` and `release` private methods or I would move the code completely to `__enter__` and `__exit__` methods, because it is risky to lock file without `with` statement.

virtwho/lock.py
```diff
@@ -0,0 +1,108 @@
...
+    def __init__(self, filename):
+        self.lock_file = filename
+
+    def acquire(self):
+        open_mode = os.O_RDWR | os.O_CREAT | os.O_TRUNC
+        fd = os.open(self.lock_file, open_mode)
+        lock_file_fd = None
+
+        timeout = 5.0
```
jirihnidek: I'm not sure about this timeout value. :-) I used 5 seconds in my example, because human being can recognize this time delay. :-) Isn't five seconds for writing to file too much?

virtwho/lock.py
```diff
@@ -0,0 +1,108 @@
...
+            print("Unable to create pid file: %s" % str(e), file=sys.stderr)
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        try:
+            os.remove(self.filename)
+        except Exception:
+            pass
+
+class FileLock(object):
+    def __init__(self, filename):
```
jirihnidek: I would add another argument `timeout` with default value `None`. When default value is used, then I would set `self.timeout` to `self.TIMEOUT`. The class timeout TIMEOUT should be specified above.

virtwho/executor.py
```diff
@@ -68,8 +74,41 @@ def _create_virt_backends(self):
                 self.logger.error('Unable to use configuration "%s": %s', name, str(e))
                 continue
             virts.append(virt)
+        self._init_run_status(config_names)
         return virts
 
+    def _init_run_status(self, config_names=[]):
+        # ensure that there is an entry in the status file for each config. We don't care if they work, we
+        # need to record their existence
+        file_lock = FileLock(STATUS_LOCK)
```
jirihnidek: I would use `with FileLock(STATUS_LOCK) as file_lock` here to be sure that the lock is released, when something unexpected happen. I can see `os.makedirs()` and `json.dump()` between `file_lock.acquire()` and `file_lock.release()`. And you never know. Somebody could add something else in the future.

virtwho/lock.py
```diff
@@ -0,0 +1,108 @@
...
+class PIDLock(object):
+    def __init__(self, filename):
+        self.filename = filename
+
+    def is_locked(self):
+        try:
+            with open(self.filename, "r") as f:
+                pid = int(f.read().strip())
+            try:
+                os.kill(pid, 0)
```
jirihnidek, line 46: I would add here comment explaining that `os.kill(pid, 0)` doesn't kill any process, but it is trick that checks if process with given `PID` still exists.

jirihnidek (changes requested):
Thanks for update of the PR. I have few small recommendations.

jirihnidek (approved):
Thanks for update. It works as expected. :+1:

PR #315, opened 2021-06-22
Status execution

wottop:
Use the -s/--status option to run virt who in status mode. It will not generate the normal logging output to the screen.
If you add the -d/--debug option, then the normal logging will return

You will see a summary of the results with just -s. If you add the option -j/--json, then you will see an expanded, json version of the results in more detail.

Note that the integration of the run-time data and the lookup of the job status will be in the next card.

Let me know if you need some config files for smoke testing.

### DIFF
--- a/tests/test_fake.py
+++ b/tests/test_fake.py
@@ -25,10 +25,13 @@ from tempfile import mkdtemp
 import shutil
 
 from base import TestBase
-
+from mock import Mock, ANY
 from virtwho.config import DestinationToSourceMapper, init_config
-from virtwho.virt import Virt, Hypervisor, VirtError
+from virtwho.virt import Virt, Hypervisor, StatusReport
 from virtwho.virt.fakevirt import FakeVirt
+from threading import Event
+from six.moves.queue import Queue
+
 
 
 HYPERVISOR_JSON = """
@@ -165,3 +168,38 @@ file=%s
         effective_config = init_config({}, config_dir=self.config_dir)
         # This is an invalid case, the config section that is invalid should have been dropped
         self.assertNotIn('test', effective_config)
+
+
+    def test_staus(self):
+
+        with open(self.hypervisor_file, "w") as f:
+            f.write(NON_HYPERVISOR_JSON)
+
+        with open(self.config_file, "w") as f:
+            f.write(f"""
+[test]
+type=fake
+is_hypervisor=false
+file={self.hypervisor_file}""")
+
+        effective_config = init_config({}, config_dir=self.config_dir)
+        manager = DestinationToSourceMapper(effective_config)
+        self.assertEqual(len(manager.configs), 1)
+        self.fake = Virt.from_config(self.logger, manager.configs[0][1], None)
+
+        self.fake.status = True
+        self.fake._send_data = Mock()
+        self.run_once()
+
+        self.fake._send_data.assert_called_once_with(data_to_send=ANY)
+        self.assertTrue(isinstance(self.fake._send_data.mock_calls[0].kwargs['data_to_send'], StatusReport))
+        self.assertEqual(self.fake._send_data.mock_calls[0].kwargs['data_to_send'].data['source']['server'], None)
+
+    def run_once(self, queue=None):
+        """Run fake in oneshot mode"""
+        self.fake._oneshot = True
+        self.fake.dest = queue or Queue()
+        self.fake._terminate_event = Event()
+        self.fake._oneshot = True
+        self.fake._interval = 0
+        self.fake._run()
--- a/tests/test_libvirtd.py
+++ b/tests/test_libvirtd.py
@@ -26,7 +26,7 @@ from mock import patch, ANY, Mock
 from virtwho import DefaultInterval
 from virtwho.virt.libvirtd.libvirtd import LibvirtdConfigSection
 from virtwho.datastore import Datastore
-from virtwho.virt import Virt, VirtError
+from virtwho.virt import Virt, VirtError, StatusReport
 
 
 def raiseLibvirtError(*args, **kwargs):
@@ -65,6 +65,27 @@ class TestLibvirtd(TestBase):
         virt.return_value.getVersion.return_value = "VERSION 1337"
         self.run_virt(config)
         virt.assert_called_with("")
+
+
+    @patch('libvirt.openReadOnly')
+    def test_read_status(self, virt):
+        config = self.create_config('test', None, type='libvirt')
+        virt.return_value.getCapabilities.return_value = LIBVIRT_CAPABILITIES_XML
+        virt.return_value.getType.return_value = "LIBVIRT_TYPE"
+        virt.return_value.getVersion.return_value = "VERSION 1337"
+        v = Virt.from_config(self.logger, config, Datastore(),
+                             interval=DefaultInterval)
+        v._terminate_event = Event()
+        v._interval = 3600
+        v._oneshot = True
+        v._createEventLoop = Mock()
+        v.status = True
+        v._send_data = Mock()
+        v._run()
+
+        v._send_data.assert_called_once_with(data_to_send=ANY)
+        self.assertTrue(isinstance(v._send_data.mock_calls[0].kwargs['data_to_send'], StatusReport))
+        self.assertEqual(v._send_data.mock_calls[0].kwargs['data_to_send'].data['source']['server'], None)
 
     @patch('libvirt.openReadOnly')
     def test_read_fail(self, virt):
--- a/virtwho/executor.py
+++ b/virtwho/executor.py
@@ -63,7 +63,8 @@ class Executor(object):
                 virt = Virt.from_config(self.logger, config, self.datastore,
                                         terminate_event=self.terminate_event,
                                         interval=self.options[VW_GLOBAL]['interval'],
-                                        oneshot=self.options[VW_GLOBAL]['oneshot'])
+                                        oneshot=self.options[VW_GLOBAL]['oneshot'],
+                                        status=self.options[VW_GLOBAL]['status'])
             except Exception as e:
                 self.logger.error('Unable to use configuration "%s": %s', name, str(e))
                 continue
@@ -93,7 +94,8 @@ class Executor(object):
                               source=self.datastore, dest=manager,
                               terminate_event=self.terminate_event,
                               interval=self.options[VW_GLOBAL]['interval'],
-                              oneshot=self.options[VW_GLOBAL]['oneshot'])
+                              oneshot=self.options[VW_GLOBAL]['oneshot'],
+                              status=self.options[VW_GLOBAL]['status'])
             dests.append(dest)
         return dests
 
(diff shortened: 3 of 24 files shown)


jirihnidek:
Result of quick testing. It works pretty well. :+1: I will do more testing today and I will also do full review of code.

virtwho/executor.py
```diff
@@ -181,6 +183,17 @@ def run_oneshot(self):
             thread.start()
 
         Executor.wait_on_threads(self.destinations)
+        if self.options[VW_GLOBAL]['status']:
+            output = {}
+            for source in self.dest_to_source_mapper.sources:
+                try:
+                    report = self.datastore.get(source)
```
jirihnidek, line 189: I guess that only this line`report = self.datastore.get(source)` is reason for try-except. I would move following two lines to `else` block. It is good practice to have only one line between try-except.

virtwho/executor.py
```diff
@@ -181,6 +183,17 @@ def run_oneshot(self):
...
+                    config = report.config
+                    output[config.name] = report
```
jirihnidek: This could be simplified to `output[report.config.name] = report`

virtwho/main.py
```diff
@@ -151,6 +144,16 @@ def main():
         logger.error(str(e))
         exit(1, "virt-who can't be started: %s" % str(e))
 
+    lock = PIDLock(PIDFILE)
+    if lock.is_locked()\
+            and not executor.options[VW_GLOBAL]['oneshot'] \
+            and not executor.options[VW_GLOBAL]['status'] \
+            and not executor.options[VW_GLOBAL]['print']:
```
jirihnidek: Maybe I would change order of checks and I would add `lock.is_locked()` to the end, because it is really not necessary to try to read lock file and check if the process is still running, when you use e.g. `--status` option.

virtwho/main.py
```diff
@@ -233,6 +240,42 @@ def _main(executor):
...
+def produce_status_output(result):
+        output = ''
+        if not executor.options[VW_GLOBAL]['json']:
+            output += ("+-------------------------------------------+\n")
+            output += ("           Configuration Status\n")
+            output += ("+-------------------------------------------+\n")
+            for config, report in result.items():
+                output += f"Configuration Name: {config}\n"
+                if 'message' in report.data['source']:
+                    output += f"Source Status: \033[1;31m{report.data['source']['status_string']}\033[0;0m\n"
```
jirihnidek: I propose to create some constants like `GREEN= \033[1;31m` and use them, because the readability of the current code is not the best and it will be easier to copy-paste code with constants. ;-)

tests/test_fake.py
```diff
@@ -165,3 +168,39 @@ def test_read_non_hypervisor_from_hypervisor(self):
...
+        with open(self.hypervisor_file, "w") as f:
+            f.write(NON_HYPERVISOR_JSON)
+
+        with open(self.config_file, "w") as f:
+            f.write("""
+[test]
+type=fake
+is_hypervisor=false
+file=%s
+""" % self.hypervisor_file)
```
jirihnidek: You can use f-string with `f"""{self.hypervisor_file}"""` too.

virtwho/executor.py
```diff
@@ -181,6 +183,17 @@ def run_oneshot(self):
...
+                except KeyError:
+                    self.logger.info('Unable to retrieve report for source '
+                                     '\"%s\" for printing' % source)
```
jirihnidek: I would use f-string here too.

virtwho/virt/virt.py
```diff
@@ -316,9 +316,47 @@ def hash(self):
...
+        self._destination_message = ''
+        self._destination = ''
+
+    @property
+    def data(self):
+        # make the components into dictionary
+        data = {"source":{"server": self.config['server'] if 'server' in self.config else 'none',
+                "status_string": 'success' if len(self._source_message) == 0 else 'failure'},
+                "destination":{"server": self._destination,
+                "status_string": 'success' if len(self._destination_message) == 0 else 'failure'}}
```
jirihnidek: Can you please reformat this little bit. It is hard to read and understand.

virtwho/virt/virt.py
```diff
@@ -316,9 +316,47 @@ def hash(self):
...
+        if len(self._source_message) > 0:
+            data['source']['message'] = self._source_message
+        if len(self._destination_message) > 0:
+            data['destination']['message'] = self._destination_message
```
jirihnidek: It is interesting that you mix two approaches. One big assignment on line 333 and then you extend the dictionary here. The question is why? I miss probably miss something.

jirihnidek (changes requested):
Thanks for PR. It works as expected, when everything (candlepin server, ESX, etc.) works as expected. When I tried to use some configuration files with wrong name of candlepin server, then I received following traceback:

```
[root@thinkpad-t580 virt-who]# ./virt-who --status --config /root/virt-who.conf.d/local_libvirt_wrong_destination.conf 
Exception in thread Thread-3:
  File "/usr/lib64/python3.9/threading.py", line 954, in _bootstrap_inner
    self.run()
  File "/home/jhnidek/githup/candlepin/virt-who/virtwho/virt/virt.py", line 475, in run
    self._send_data(data_to_send=report)
  File "/home/jhnidek/githup/candlepin/virt-who/virtwho/virt/virt.py", line 665, in _send_data
    valid = self.dest.hypervisorHeartbeat(self.config, options=self.options)
  File "/home/jhnidek/githup/candlepin/virt-who/virtwho/manager/subscriptionmanager/subscriptionmanager.py", line 270, in hypervisorHeartbeat
    connection = self._connect(config)
  File "/home/jhnidek/githup/candlepin/virt-who/virtwho/manager/subscriptionmanager/subscriptionmanager.py", line 167, in _connect
    if not self.connection.ping()['result']:
  File "/usr/lib64/python3.9/site-packages/rhsm/connection.py", line 1044, in ping
    return self.conn.request_get("/status/")
  File "/usr/lib64/python3.9/site-packages/rhsm/connection.py", line 914, in request_get
    return self._request("GET", method, headers=headers, cert_key_pairs=cert_key_pairs)
  File "/usr/lib64/python3.9/site-packages/rhsm/connection.py", line 939, in _request
    result = super(Restlib, self)._request(request_type, method,
  File "/usr/lib64/python3.9/site-packages/rhsm/connection.py", line 736, in _request
    conn.request(request_type, handler, body=body, headers=final_headers)
  File "/usr/lib64/python3.9/http/client.py", line 1253, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/usr/lib64/python3.9/http/client.py", line 1299, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
  File "/usr/lib64/python3.9/http/client.py", line 1248, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "/usr/lib64/python3.9/http/client.py", line 1008, in _send_output
    self.send(msg)
  File "/usr/lib64/python3.9/http/client.py", line 948, in send
    self.connect()
  File "/usr/lib64/python3.9/http/client.py", line 1415, in connect
    super().connect()
  File "/usr/lib64/python3.9/http/client.py", line 919, in connect
    self.sock = self._create_connection(
  File "/usr/lib64/python3.9/socket.py", line 822, in create_connection
    for res in getaddrinfo(host, port, 0, SOCK_STREAM):
  File "/usr/lib64/python3.9/socket.py", line 953, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
socket.gaierror: [Errno -2] Name or service not known

^C[root@thinkpad-t580 virt-who]# ca/root/virt-who.conf.d/local_libvirt_wrong_destination.confnf
[local_libvirt_wrong]
type=libvirt
server=qemu:///system
owner=admin

rhsm_hostname=foo.bar.do-not-exist.org
rhsm_port=8443
rhsm_username=admin
rhsm_password=admin
rhsm_org=admin
rhsm_prefix=/candlepin
rhsm_insecure=1
```

I also left some comments in the code. I also noticed that you use f-strings in some parts :+1:

virtwho/virt/virt.py
wottop: I didn't want the element label to appear if there is no message, but it's not a big deal. updated.

tests/test_libvirtd.py
```diff
@@ -66,6 +66,27 @@ def test_read(self, virt):
...
+        v._interval = 3600
+        v._oneshot = True
+        v._createEventLoop = Mock()
+        v.status = True
+        v._send_data = Mock()
+        v._run()
+
+        v._send_data.assert_called_once_with(data_to_send=ANY)
+        self.assertTrue(isinstance(v._send_data.mock_calls[0].kwargs['data_to_send'], StatusReport))
+        self.assertEqual(v._send_data.mock_calls[0].kwargs['data_to_send'].data['source']['server'], 'none')
```
jirihnidek: I wonder why it has to equal to `'none'` string for libvirtd?

tests/test_fake.py
```diff
@@ -165,3 +168,38 @@ def test_read_non_hypervisor_from_hypervisor(self):
...
+        self.assertEqual(len(manager.configs), 1)
+        self.fake = Virt.from_config(self.logger, manager.configs[0][1], None)
+
+        self.fake.status = True
+        self.fake._send_data = Mock()
+        self.run_once()
+
+        self.fake._send_data.assert_called_once_with(data_to_send=ANY)
+        self.assertTrue(isinstance(self.fake._send_data.mock_calls[0].kwargs['data_to_send'], StatusReport))
+        self.assertEqual(self.fake._send_data.mock_calls[0].kwargs['data_to_send'].data['source']['server'], 'none')
```
jirihnidek: Why it has to be equal to `'none'` in this case?

jirihnidek: Oh, I can see it. `type=fake` does not have any server.

tests/test_libvirtd.py
jirihnidek: I can see it now. The libvirtd does not have any server in this case.

virtwho/virt/virt.py
```diff
@@ -316,9 +315,50 @@ def hash(self):
...
+        self._source_message = ''
+        self._destination_message = ''
+        self._destination = ''
+
+    @property
+    def data(self):
+        # make the components into dictionary
+        data = {
+            "source": {
+                "server": self.config['server'] if 'server' in self.config else 'none',
```
jirihnidek: I wonder why do you `'none'` string and not `None`? I'm not telling it is wrong. I just would like to know the reason.

jirihnidek (approved):
Thanks for the update of this PR. It looks OK. I have only one last question in the comment.

virtwho/virt/virt.py
wottop: It was a placeholder when I started. I think it was for display purposes, but I agree that None is proper.

jirihnidek (approved):
ACK

jirihnidek:
I don't understand why some Packit jobs are marked as failed. When I click at details, then I don't see there anything wrong. Probably some error in communications. I can also build and install RPM on my testing machine and everything works there as expected.

PR #316, opened 2021-07-07
Update the man page for the status mode

wottop:


virt-who.8
```diff
@@ -36,12 +42,17 @@ In this mode virt-who just sends the host to guest association to the server onc
...
 .TP
 3. print mode
 # virt-who -p
 
 This mode is similar to oneshot mode but the host to guest association is not send to server, but printed to standard output instead.
+.TP
+3. status mode
+# virt-who -s
+
+This mode is for configuration diagnosis. The host to guest association is not compiled and is not sent to the server.
```
cnsnyder: I'd suggest we tell the users a bit more about what is actually happening here.
As an example, how do you use this to diagnose issues with configurations?
What should a user expect to see?

cnsnyder (approved):

PR #319, opened 2021-07-16
Virt who status feature

wottop:
Each commit has been through a PR already. This is for the final merge to main.
Let me know if you see anything that we missed the first time.


jirihnidek (approved):
ACK

tlhmerry9 (approved):
This implementation makes sense to me.

PR #330, opened 2021-08-04
1985210: The guest state in mapping should be uniform with other hypervisors [8.5]

wottop:
1985213: The guest shows wrong active value "0" in mapping when it's running.


wottop:
Reference for acceptable values: 

http://man.hubwiz.com/docset/Terraform.docset/Contents/Resources/Documents/docs/providers/nutanix/d/virtual_machine.html

cnsnyder (approved):

PR #331, opened 2021-08-05
1990337: The guest state in mapping should be uniform with other hypervisors [main]

wottop:
1990338: The guest shows wrong active value "0" in mapping when it's running


cnsnyder (approved):
LGTM

PR #335, opened 2021-08-12
1989877: Status command does not reach actual credentials checking

wottop:
Affected hyperv, kubevirt, libvirtd, and rhevm

### DIFF
--- a/virtwho/virt/hyperv/hyperv.py
+++ b/virtwho/virt/hyperv/hyperv.py
@@ -625,5 +625,15 @@ class HyperV(virt.Virt):
         hypervisor = virt.Hypervisor(hypervisorId=host, name=hostname, guestIds=guests, facts=facts)
         return {'hypervisors': [hypervisor]}
 
+    def statusConfirmConnection(self):
+        '''
+        This call will confirm the credentials. The result outside
+        of that is not important in the status scenario.
+        '''
+        connection = self.connect()
+        hypervsoap = HyperVSoap(self.url, connection, self.logger)
+        hypervsoap.Invoke_GetSummaryInformation(
+            "root/virtualization/v2" if self.useNewApi else "root/virtualization")
+
     def ping(self):
         return True
--- a/virtwho/virt/kubevirt/kubevirt.py
+++ b/virtwho/virt/kubevirt/kubevirt.py
@@ -154,3 +154,10 @@ class Kubevirt(virt.Virt):
             hosts[host_name].guestIds.append(virt.Guest(guest_id, self.CONFIG_TYPE, status))
 
         return {'hypervisors': list(hosts.values())}
+
+    def statusConfirmConnection(self):
+        '''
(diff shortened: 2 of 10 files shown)


virtwho/virt/virt.py
```diff
@@ -1210,6 +1213,9 @@ def listDomains(self):
         '''
         raise NotImplementedError('This should be reimplemented in subclass')
 
+    def statusConfirmConnection(self):
```
jirihnidek, line 1216: Can you please add here some doc-string explaining purpose of this method (what should be called in subclass, etc.). It could be helpful for anybody trying to add support for another type of hypervisor.

virtwho/virt/rhevm/rhevm.py
```diff
@@ -278,6 +278,9 @@ def getHostGuestMapping(self):
 
         return {'hypervisors': list(hosts.values())}
 
+    def statusConfirmConnection(self):
```
jirihnidek, line 281: Can you please add doc-string here?

virtwho/virt/libvirtd/libvirtd.py
```diff
@@ -441,3 +444,7 @@ def _getHostGuestMapping(self):
                           facts=facts)
         mapping['hypervisors'].append(host)
         return mapping
+
+    def statusConfirmConnection(self):
```
jirihnidek, line 448: Can you please add some doc-string here.

virtwho/virt/kubevirt/kubevirt.py
```diff
@@ -154,3 +154,6 @@ def getHostGuestMapping(self):
             hosts[host_name].guestIds.append(virt.Guest(guest_id, self.CONFIG_TYPE, status))
 
         return {'hypervisors': list(hosts.values())}
+
+    def statusConfirmConnection(self):
```
jirihnidek, line 158: doc-string please.

virtwho/virt/hyperv/hyperv.py
```diff
@@ -625,5 +625,11 @@ def getHostGuestMapping(self):
         hypervisor = virt.Hypervisor(hypervisorId=host, name=hostname, guestIds=guests, facts=facts)
         return {'hypervisors': [hypervisor]}
 
+    def statusConfirmConnection(self):
```
jirihnidek, line 628: doc-string please.

jirihnidek (changes requested):
Overall, I like this PR. :+1: 

I miss some doc-strings. When you see methods `statusConfirmConnection()` implemented in sub-classes, then it could be hard to understand, why it tries to do, what it does, when you do not know the context.

Can you please add at least one unit test for this case too?

jirihnidek:
I can also see failing unit tests on CI, but it is probably unrelated to the changes in the PR.

jirihnidek (approved):
Thanks for update. :+1:

PR #336, opened 2021-08-24
hyperv: Stop using NTLM which requires MD4 (OpenSSL 3.0)

cnsnyder:
OpenSSL 3.0 dropped support for MD4 in all but it's legacy mode.
This is a minimal change to stop using NTLM and rely only on
Basic auth. Please note that this is a simiar level of security
as the other virtualization providers.


wottop (changes requested):
Please reference https://bugzilla.redhat.com/show_bug.cgi?id=1981249 in the commit message.

wottop:
ack

wottop (approved):

PR #337, opened 2021-08-24
1985929: Print logging to stderr when status and debug are both used

wottop:


virtwho/log.py
```diff
@@ -282,6 +284,8 @@ def get_file_handler(cls, name, config=None):
     @classmethod
     def get_stream_handler(cls, name):
         if cls._stream_handler is not None:
+            if cls._debug and cls._status:
```
jirihnidek, line 287: Why is stream set to `stderr` here, when the `cls._stream_handler` already exists? This block code is used for caching.

jirihnidek (changes requested):
It looks like that this PR does not have any effect. When debug mode is enabled, then logging is already printed to `stderr` and it can be filtered out using `2> /dev/null`. Or do I miss anything? In case I miss something, can you please add unit tests for this case?

wottop:
You are right. I just followed the BZ and didn't confirm. Closing.

PR #338, opened 2021-08-26
Add stylish

m-horky:
Card ID: ENT-4004

This set of commits does the following:
- Add flake8 to Makefile 'stylish' command
- Add Jenkins 'stylish' job
- Format files so they are 'flake8-compliant' (two files have been converted to 4-space indentation)
- Format test files

Please note that some tests had to be slightly altered, mostly because they were defining variables that weren't used in the test (these variables were removed). The only difference being `test_config_section_virt.py` in `test_validate_wrong_values_in_encrypted_password` where two lines were added to restore the `password` variable:
https://github.com/candlepin/virt-who/blob/d38b0f40e2d895b42ef33182133bdbe9090575e9/tests/test_config_section_virt.py#L400-L404


jirihnidek (approved):
LGTM. It's big PR. So I'm only adding ACK and I would like to see another ACK from at least one more developer.

virtwho/virt/virt.py
```diff
@@ -1120,13 +1124,13 @@ def __init__(self, logger, config, dest, terminate_event=None,
...
-        import virtwho.virt.fakevirt  # flake8: noqa
-        import virtwho.virt.kubevirt  # flake8: noqa
-        import virtwho.virt.ahv       # flake8: noqa
+        import virtwho.virt.libvirtd
+        import virtwho.virt.esx
+        import virtwho.virt.rhevm
+        import virtwho.virt.hyperv
+        import virtwho.virt.fakevirt
+        import virtwho.virt.kubevirt
+        import virtwho.virt.ahv  # noqa: F401
```
wottop, line 1136: What are we missing here? Is there a separate card for it?

m-horky: These are needed just below for the `from_config()`. Is there a way to do this without importing them like this? I only did the minimum to keep flake8 happy, so it's not upset about imports not being on top of the file.

wottop: ok

m-horky:
I did a rebase, the suds change affected the same portion of code (the imports in `virt.py`).

jirihnidek (approved):
I checked all changes again and it looks good to me. :+1:

PR #339, opened 2021-08-30
Cannot use ESX where python3-suds is not available

wottop:


cnsnyder:
Is there any way we could give the user some sort of message about this special case?
I guess I'm just wondering which is worse:

Outputting log messages which seem to indicate that virt-who has no idea what ESX/VMware is

or

Printing out something potentially more scary about knowing what ESX is but not being able to communicate with it due to missing dependencies.

I guess I just want to be as transparent as possible while hopefully remaining actionable. That last part is tough in this case though as there really isn't anything supported users could do.

cnsnyder (approved):
LGTM! Thank you very much for getting that doc added!

PR #344, opened 2021-09-27
1889363: Update dependency between libvirtd and virt-who

wottop:
This addition is based on this comment in the BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1889363#c12

### DIFF
--- a/virt-who.service
+++ b/virt-who.service
@@ -1,6 +1,7 @@
 [Unit]
 Description=Daemon for reporting virtual guest IDs to subscription-manager
 After=libvirtd.service network.target
+Wants=libvirtd-ro.socket
 
 [Service]
 Type=notify


virt-who.service
```diff
@@ -1,6 +1,7 @@
 [Unit]
 Description=Daemon for reporting virtual guest IDs to subscription-manager
 After=libvirtd.service network.target
+Wants=libvirt-ro.socket
```
jirihnidek: I don't have anything like `libvirt-ro.socket` on my systems. I have only `libvirtd-ro.socket` on my systems.

jirihnidek (changes requested):
I guess that there is typo in your commit.

jirihnidek (approved):
I can confirm that `systemct start virt-who.service` automatically starts `libvirtd-ro.socket`, when libvirt is installed. When `libvirt` is not installed at all, then virt-who.service is started without any problem.

PR #345, opened 2021-10-01
Remove unused NTLM code as only basic auth is used [ENT-4257]

wottop:
NTLM is no longer an authentication protocol that we can use because of changes in the openssl package.


cnsnyder (approved):
LGTM!

PR #346, opened 2021-10-07
Only query the VM related taks in AHV hypervisors.

aeibagi:
This change will only query the VM related tasks from AHV
clusters during continous mode to reduce the laod in returning
the tasks in the AHV task manager. A new wait time flag is also
added to cause delay in querying for the tasks in the continous
mode.

PR #347, opened 2021-10-18
This change will only query the VM related tasks from AHV

aeibagi:
clusters during continous mode to reduce the laod in returning
the tasks in the AHV task manager. A new wait time flag is also
added to cause delay in querying for the tasks in the continous
mode.


aeibagi:
Pull request against 1.30 branch. 
The original Pull request is here:  https://github.com/candlepin/virt-who/pull/346

wottop:
ack

PR #348, opened 2021-10-29
Document rhsm_insecure config option

danifernandezs:
Added to man (5) the following configuration option

- rhsm_insecure: Boolean to avoid the TLS verification at the connection stablishment


virt-who-config.5
```diff
@@ -116,6 +116,12 @@ Alternative to the rhsm_proxy_password option; encrypted password generated by t
 \fBrhsm_no_proxy\fR
 Optional proxy settings for the Subscription Asset Manager or Satellite 6 server to use in place of the no proxy filter defined in the system's rhsm.conf.
 .TP
+\fBrhsm_certificate\fR
```
wottop: This is not used in virt-who. It is a setting in subscription-manager, but virt-who has never passed it from its own configuration.

wottop (changes requested):

virt-who-config.5
danifernandezs: when I tried, I saw that really doing nothing, that explain the root cause, hahaha

sorry, but I only can to read a few code lines and try to test if my interpretation of the readed code works as my needs

thanks for the aclaration

ptoscano:
Hi @danifernandezs,

the change seems OK from my POV; can you please
- improve the commit message for the change, as "docs:feat: rhsm_insecureinstead insecure rhsm.conf" is a bit too terse, and lacks whitespaces
- rebase on top of the current `main` branch

Thanks! I'll defer to @wottop for the final review.

wottop:
@danifernandezs can we get the issues raised by @ptoscano corrected?
I'd love to get this into the application.

danifernandezs:
Hi @wottop,

I have completed the rebase and changed the commit message. I hope I have done it correctly, as my knowledge of git is limited to just committing and pushing. In fact, I had to perform a main rebase, an amend, and a push with --force, as it was not possible to do it any other way.

ptoscano (approved):
LGTM, thanks!

PR #351, opened 2021-11-30
Update for deprecation of MutableMapping [main]

wottop:
The collections.MutableMapping was deprecated in Python 3.3
and the new collections.abc.MutableMapping should be used.
The old API will is completely removed in Python 3.10.


jirihnidek (approved):
ACK

PR #352, opened 2021-12-08
2018052: Fix virt-who -s, when Hyper-V with new API is used

jirihnidek:
* BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2018052
* ENT-4533
* When old Hyper-V API was used, then --status worked as expected,
  but virt-who wasn't able to resurect gracefully, when only
  new Hyper-V API was supported.
* virt-who --status tries to use old API and if it is not
  available, then it tries to use new API
* Fixed two unit tests and indentation in some unit tests


wottop:
ack

PR #354, opened 2021-12-15
2000019: Convert non-latin1 username/password to bytes

jirihnidek:
* BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2000019
* Card ID: ENT-4307
* The requests package does not allow non latin1 strings to be
  used for authentication. It has to be converted to bytes.
* It wasn't possible to concert username/password directly
  to bytes in Ahv, because username/password setters of Virt
  convert these values to strings
* Added few unit tests for this case


cnsnyder:
There is a typo in the commit message "concert". I think "convert" was meant.

cnsnyder (approved):

PR #357, opened 2022-01-06
1986749: rename configuration entry to show ahv specificity [1.30]

wottop:


virtwho/virt/ahv/ahv_interface.py
```diff
@@ -310,7 +310,7 @@ def _send(self, method, url, **kwargs):
     for ii in range(retry_count):
       try:
         response = func(url, **kwargs)
-        if self._internal_debug:
+        if self._ahv_internal_debug:
```
jirihnidek, line 313: I see here one issue and it is not related directly with this change. So, I would rather try to fix it in another PR.

OK, what is the issue? When the `ahv_internal_debug` is set on in configuration file and it is not possible to authenticate, because you have wrong credentials, then decoding of JSON document fails, because server response with html document.

jirihnidek (approved):
ACK

PR #360, opened 2022-01-26
1996942: Use cluster name instead of UUID for fabric consistency

wottop:


virtwho/virt/ahv/ahv_interface.py
```diff
@@ -596,20 +596,27 @@ def get_hosts_uuid(self, version):
...
-            return host_info['cluster_reference']['uuid']
+            this_uuid = host_info['cluster_reference']['uuid']
 
-    def get_ahv_cluster_uuids(self, version):
+        for uuid, name in cluster_ids:
+            if this_uuid == uuid:
+                return name
+        return this_uuid
+
+    def get_ahv_cluster_uuid_map(self, version):
```
jirihnidek, line 622: I thinks that doc string should be changed according change in the code. E.g the method returns list of tuples.

virtwho/virt/ahv/ahv_interface.py
```diff
@@ -596,20 +596,27 @@ def get_hosts_uuid(self, version):
...
+            this_uuid = host_info['cluster_uuid']
         elif 'cluster_reference' in host_info:
-            return host_info['cluster_reference']['uuid']
+            this_uuid = host_info['cluster_reference']['uuid']
 
-    def get_ahv_cluster_uuids(self, version):
+        for uuid, name in cluster_ids:
+            if this_uuid == uuid:
+                return name
+        return this_uuid
```
jirihnidek, line 620: I would do further refactoring of code. It is little bit confusing that you return here `this_uuid`, when the method should return name. It looks like that the method could in theory return UUID. If method can return None, then it should be documented in doc string.

virtwho/virt/ahv/ahv.py
```diff
@@ -114,6 +114,8 @@ def getHostGuestMapping(self):
         mapping = {'hypervisors': []}
 
         host_uvm_map = self._interface.build_host_to_uvm_map(self.version)
+        if host_uvm_map:
+            cluster_ids = self._interface.get_ahv_cluster_uuid_map(self.version)
```
jirihnidek: I would rename this method to `cluster_uuid_map` to emphasize that it is map of uuids.

virtwho/virt/ahv/ahv.py
```diff
@@ -149,15 +151,15 @@ def getHostGuestMapping(self):
             else:
                 self.logger.debug("Host '%s' doesn't have any vms", host_uuid)
 
-            cluster_uuid = self._interface.get_host_cluster_uuid(host)
+            cluster_name = self._interface.get_host_cluster_name(host, cluster_ids)
```
jirihnidek: Is it really possible that the cluster_name is `None`? What should we do in this case? If it is possible then we probably should not add the `hypervisor.cluster` to the dictionary of facts. Thoughts?

jirihnidek (changes requested):
I can confirm that the PR works as expected. :+1:  I have few requests in the code and suggestion how to improve readability of code.

virtwho/virt/ahv/ahv.py
wottop: Perhaps something like UUID:[uuid]. It has a cluster with an id, but no textual name.

jirihnidek: When the cluster has only UUID but it does not have name, then it is OK to print some debug message about it and use UUID for `hypervisor.cluster`.

When the `cluster_name` is None, then `hypervisor.cluster` should not be added to dictionary.

wottop:
Once we agree on the code, I'll move the changes to the other PR.

jirihnidek (approved):
LGTM, Thanks

PR #363, opened 2022-01-26
1985258: Move progress bar to stderr to keep stdout clean

wottop:
Stdout is used for redirection to create json for the fake fabric.
The output on the screen will appear the same, but the progress
 bar will not be included in the redirect.


jirihnidek (approved):
LGTM

PR #364, opened 2022-01-27
2000415: Nutanix config needs to handle bad server value

wottop:


tests/test_ahv.py
```diff
@@ -413,16 +413,6 @@ def test_validate_ahv_PE_config(self):
         result = self.ahv_config.validate()
         self.assertEqual(len(result), 0)
 
-    def test_validate_ahv_invalid_server_ip(self):
```
cnsnyder, line 416: Why is this test being removed? (Sorry it is not apparent to me for some reason)

tests/test_config_section_ahv.py
```diff
@@ -0,0 +1,70 @@
+# -*- coding: utf-8 -*-
+from __future__ import print_function
```
cnsnyder: Do we need this against main?

virtwho/virt/ahv/ahv.py
```diff
@@ -276,11 +276,16 @@ def _validate_server(self, key):
             Socket error is returned in case of an invalid ip.
         """
         error = super(AhvConfigSection, self)._validate_server(key)
-        try:
-            ip = self._values[key]
-            socket.inet_aton(ip)
-        except socket.error:
-            error = 'Invalid server IP address provided'
+        if error is None:
```
cnsnyder, line 279: Any chance this could end up being an empty list or empty tuple at this point (anything else false-like but not None)?

cnsnyder: From looking more at the VirtConfigSection._validate_server() method, it seems checking for None only works.

cnsnyder (approved):
Looks good to me! One minor non-nacking suggestion: if we don't need the __future__ import any more, I think we should not include it in new bits as we go forward.

tests/test_config_section_ahv.py
wottop: It is consistent with the other modules. Perhaps we make it uniform and then remove them all at the same time.

virtwho/virt/ahv/ahv.py
wottop: it is consistent with the other _validate_server override methods.

PR #365, opened 2022-01-27
2000423: Nutanix config needs to handle bad server value

wottop:


cnsnyder (approved):
These are the same (less the future import) changes as were made against main for #364 ! :+1:

PR #368, opened 2022-01-31
Addition of LGPLv3 license file from suds; ENT-4351

wottop:


cnsnyder:
Looks good (and builds fine) to me! Where did you get the license from? (Just making sure we have the right one for the source we have)

wottop:
Added URL of the original license file to commit message.

PR #375, opened 2022-04-13
2054504: Use usedforsecurity=False for md5() calls to make suds work on FIPS enabled systems

ptoscano:
Backport https://github.com/suds-community/suds/pull/72 to the virt-who copy of suds.


jirihnidek (approved):
ACK

PR #376, opened 2022-05-30
Fix virt-who for AHV

need349:
We're using Redhat Satellite 6.10.5 with Nutanix AHV as described in article https://access.redhat.com/articles/6216971 with an instance of Nutanix Prism Central for reporting the host/guest-mapping with virt-who.
It seems, that virt-who always reports the same VM's as the filter with the Nutanix REST API v3 seems not get applied.

Virt-Who Version: 1.30.12-1.el8 (via RHEL Repositorys)
OS: Red Hat Enterprise Linux 8
Prism Central: 2022.4

Since the last update of virt-who, we had problems, that virt-who just reports a small part of vms and not all.
I've put a little bit of research and some cups of coffee in it and found out, that the function make_rest_call in ahv_interface.py doesn't applies the filter length and offset correctly. So virt-who always asks the Prism Central API for the first 20 vms over and over again without any body for filtering. I hope, that it's understandable with the screen attachemend and the empty json body ({}). The problem seems to be also in the latest version of virt-who, so I'm running the pull request on 1.31.22.

I've fixed that by adjusting the ahv_interface.py. Also I've increased the length of the request chunks from 20 to 100 (max length from nutanix api is 500), to improve the virt-who performance a little bit. Otherwise a run with virt-who is with 1800 vms on nutanix is extremely slow and takes over 30 minutes per run.

This is the very first pull request on this project, so if you miss anything, just let me know!

grafik


virtwho/virt/ahv/ahv_interface.py
```diff
@@ -514,14 +514,17 @@ def get_vms_uuid(self, version):
         self._logger.info("Getting the list of available vms")
         is_pc = True if version == 'v3' else False
         vm_uuid_list = []
-        length = 0
+        length = 100
```
jirihnidek: Could you please add constants 100 as a class attribute (something like `NUM_OF_REQUESTED_VMS`) ?

virtwho/virt/ahv/ahv_interface.py
```diff
@@ -514,14 +514,17 @@ def get_vms_uuid(self, version):
...
         offset = 0
```
jirihnidek, line 519: Could you please rename this variable to something like `initial_offset`?

jirihnidek (changes requested):
Thanks for the PR. Overall it looks OK, but I have to test it more, because currently I have access to hypervisor only with one guest and I'm not admin of this machine.

I have small requests in comments of the code ATM.

Can you please add at least one unit test for this case? If you don't know how to add it, then it is OK and we can add unit tests later.

virtwho/virt/ahv/ahv_interface.py
```diff
@@ -514,14 +514,17 @@ def get_vms_uuid(self, version):
...
         total_matches = 0
         count = 1
         current = 0
         (url, cmd_method) = self.get_diff_ver_url_and_method(
             cmd_key='list_vms', intf_version=version)
-        res = self.make_rest_call(method=cmd_method, uri=url)
+        body = {
+            'length': length
+        }
+        res = self.make_rest_call(method=cmd_method, uri=url, json=body)
```
jirihnidek: It doesn't make sense to set body, when version prism v2 REST API is used. It does not make sense to set body for GET method

https://www.nutanix.dev/api_references/prism-v2-0/#/b3A6MjU1ODkwNzk-get-a-list-of-virtual-machines

virtwho/virt/ahv/ahv_interface.py
```diff
@@ -514,14 +514,17 @@ def get_vms_uuid(self, version):
...
+        length = 100
         offset = 0
         total_matches = 0
         count = 1
         current = 0
         (url, cmd_method) = self.get_diff_ver_url_and_method(
             cmd_key='list_vms', intf_version=version)
-        res = self.make_rest_call(method=cmd_method, uri=url)
+        body = {
+            'length': length
```
jirihnidek: I would set `offset` (equal to 0) here too due to consistency.

jirihnidek (changes requested):
I verified that proposed change works as expected. I have more requests.

need349:
Thanks for the review! I've just fixed the comments, just let me know if I miss anything.
I'm also working on a unit test for this case, just give me some days to work that out..

virtwho/virt/ahv/ahv_interface.py
```diff
@@ -514,14 +514,20 @@ def get_vms_uuid(self, version):
         self._logger.info("Getting the list of available vms")
         is_pc = True if version == 'v3' else False
         vm_uuid_list = []
-        length = 0
```
jirihnidek, line 517: You cannot do that, because `length` does not exists on line 545 and it causes failures.

virtwho/virt/ahv/ahv_interface.py
```diff
@@ -514,14 +514,20 @@ def get_vms_uuid(self, version):
...
-        offset = 0
```
jirihnidek, line 518: I suggested to rename this variable to `initial_offset` not removing it. This removal is causing failures at line 555.

jirihnidek (changes requested):
Hi, thanks for updates, but your changes made it worse. Your changes cause failures. You can see it, when you e.g. run unit tests. You can run unit tests using `py.test` command.

jirihnidek:
BTW: commits should be squashed into one commit in this case.

jirihnidek:
The code also fails with following traceback during testing `./virt-who -d -o -c /path/to/nutanix.conf` and nothing is reported to candlepin server (I use setup with only two VMs):

```
2022-06-29 12:39:23,958 [virtwho.main ERROR] MainProcess(72411):Thread-2 @virt.py:run:527 - Thread 'nutanix_conf' fails with exception:
  File "/home/jhnidek/githup/candlepin/virt-who/virtwho/virt/virt.py", line 518, in run
    self._run()
  File "/home/jhnidek/githup/candlepin/virt-who/virtwho/virt/ahv/ahv.py", line 195, in _run
    assoc = self.getHostGuestMapping()
  File "/home/jhnidek/githup/candlepin/virt-who/virtwho/virt/ahv/ahv.py", line 116, in getHostGuestMapping
    host_uvm_map = self._interface.build_host_to_uvm_map(self.version)
  File "/home/jhnidek/githup/candlepin/virt-who/virtwho/virt/ahv/ahv_interface.py", line 776, in build_host_to_uvm_map
    vm_uuids = self.get_vms_uuid(version)
  File "/home/jhnidek/githup/candlepin/virt-who/virtwho/virt/ahv/ahv_interface.py", line 555, in get_vms_uuid
    'offset': offset
NameError: name 'offset' is not defined
```

need349:
Sorry, unfortunately I missed, that the variable is used for a lot of other stuff. I fixed that and ran multiple tests with Prism Central and Prism Elements and the report works as expected. If I miss something, just let me know!

wottop:
There is a stylish error:

./virtwho/virt/ahv/ahv_constants.py:27:2: W292 no newline at end of file

1     W292 no newline at end of file

jirihnidek:
Closing this PR, because #377 (containing code provided in this PR) was merged to main. Thanks again for your contribution.

PR #377, opened 2022-07-27
2091661: Nutanix: Gather information about VMs correctly

jirihnidek:
* BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2091661
* Card ID: ENT-5116
* Original PR and commits created by Robin Mazzeo
  - PR: https://github.com/candlepin/virt-who/pull/376
* When version 3 of Nutanix API was used, then it was
  possible to gather only first 20 VMs

### DIFF
--- a/virtwho/virt/ahv/ahv_interface.py
+++ b/virtwho/virt/ahv/ahv_interface.py
@@ -514,14 +514,30 @@ class AhvInterface(object):
         self._logger.info("Getting the list of available vms")
         is_pc = True if version == 'v3' else False
         vm_uuid_list = []
-        length = 0
+        length = ahv_constants.NUM_OF_REQUESTED_VMS
+        initial_offset = 0
         offset = 0
         total_matches = 0
         count = 1
         current = 0
-        (url, cmd_method) = self.get_diff_ver_url_and_method(
+        url, cmd_method = self.get_diff_ver_url_and_method(
             cmd_key='list_vms', intf_version=version)
-        res = self.make_rest_call(method=cmd_method, uri=url)
+
+        kwargs = {
+            "method": cmd_method,
+            "uri": url
+        }
+        if is_pc is True:
+            kwargs["json"] = {
+                'length': length,
+                'offset': initial_offset
+            }
+        res = self.make_rest_call(**kwargs)
+
+        if res is None:
+            self._logger.error("Unable to get list of VMs")
+            return vm_uuid_list
+
         data = res.json()
         if "metadata" in data:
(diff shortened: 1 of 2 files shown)


virtwho/virt/ahv/ahv_interface.py
```diff
@@ -557,11 +565,12 @@ def get_vms_uuid(self, version):
...
-            body['offset'] = body['offset'] + length
-            body_data = json.dumps(body, indent=4)
-            self._logger.debug('next vm list call has this body: %s' % body)
-            res = self.make_rest_call(method=cmd_method, uri=url, data=body_data)
+            if is_pc is True:
+                body['offset'] = body['offset'] + length
+                self._logger.debug('Next vm list call has this body: %s' % body)
+                res = self.make_rest_call(method=cmd_method, uri=url, json=body)
+            else:
+                res = self.make_rest_call(method=cmd_method, uri=url)
```
jirihnidek: Note: it is probably not necessary to use this for loop, when version 2.0 is used, because complete list is gathered in first REST API call outside of this for loop. I will try to fix it in different PR.

jirihnidek: I only hope that Nutanix returns complete list, when version 2.x is used. There is no limitation described in the documentation of API: https://www.nutanix.dev/api_references/prism-v2-0/#/b3A6MjU1ODkwNzg-get-a-list-of-virtual-machines On the other side documentation says that query arguments `offset` and `lengths` could be used.

virtwho/virt/ahv/ahv_interface.py
```diff
@@ -514,14 +514,22 @@ def get_vms_uuid(self, version):
...
             cmd_key='list_vms', intf_version=version)
-        res = self.make_rest_call(method=cmd_method, uri=url)
+        if is_pc is True:
+            body = {
+                'length': length,
+                'offset': initial_offset
+            }
+            res = self.make_rest_call(method=cmd_method, uri=url, json=body)
+        else:
+            res = self.make_rest_call(method=cmd_method, uri=url)
```
ptoscano: I think this can be simplified a bit, removing the duplicated `make_rest_call()` call & args, by composing its arguments as kwargs.

virtwho/virt/ahv/ahv_interface.py
ptoscano: ditto (wrt using kwargs to avoid code copying)

ptoscano (commented):
only a couple of notes for now, as this needs to be tested in a real environment

PR #378, opened 2022-07-29
 Big optimization and refactoring of Nutanix code

jirihnidek:
* Card ID: ENT-5318
* We support to versions of Nutanix REST API (version 2 and
  version 3). Each version has its own class AhvInterface2
  and AhvInterface3 now. There are small and big differences
  in the behavior of REST API endpoints for each version
  and this approach will help to develop and fix issues
  in more reliable manner.
* We gather only list of cluster, list of hosts and list
  of VMs, because these lists contains all information,
  we need to assemble host-guests mapping. It is really
  not necessary to gather again information for each
  host nor VM, because these information are already
  gathered in the list. Thus gathering of data is much
  more faster (depends on number of clusters, hosts
  and VMs)
* We keep HTTP session and we don't close connection
  after each request. It makes communication with server
  faster
* Removed progress bar
* The detection of changes on server was dropped, because
  the list of changes usually contained too many changes
  and returned list wasn't complete. I have also observed
  that there have always been some changes (not relevant
  to host-guest mapping) and it triggered gathering list
  of data anyway. We use same approach as we do for other
  hypervisors. This approach has one benefit: the
  virt-who using Nutanix can be terminated gracefully.
  Previous approach was also hard or impossible to test.
* Completely removed configuration option
  "wait_time_in_sec", because it is not used anymore.
* Removed ahv_constants, because this ".h" file is
  not pythonic. All constants were moved to classes.
* When version 3 is used, then "active" state of VM
  is set correctly.
* Added GPL license to each file
* Fixed several typos and style issues
* Fixed several unit tests
* Future work: use threads for gathering long lists
  of clusters, hosts and VMs. We don't have such
  instance for testing ATM.


jirihnidek:
Few notes.
* I haven't expected that the code is so inefficient... Trying to support two versions in the same code base just wasn't good idea.
* Example of inefficiency: The method `build_host_to_uvm_map()` tries to get list of VMs using method `get_vms_uuid()` and then it tries to get detail information for each VM from the list. This is right approach for API v2.0, because the list of VMs contains only minimal information. When version 3.x is used, then the list of VMs also contains `host_uuid`. Thus it is not necessary to gather detailed information for each VM later.
* Thus gathering of complete list of VM should be multi-threaded, when version 3.x is used
* And gathering of VM details should be multi-threaded, when version 2.x is used.

jirihnidek:
Gathering information about one cluster, one host and 22 VMs:

this branch: 6 seconds
main branch: 50 seconds

Cheers :-)

wottop:
Can you put the BZ # for the "When version 3 is used, then "active" state of VM
is set correctly." issue in the commit message? There is a parallel BZ for RHEL 8 that will need to be in 1.30.

https://bugzilla.redhat.com/show_bug.cgi?id=2122077

Otherwise LGTM

jirihnidek:
Commit and BZ updated

wottop:
ack, holding off merge until 9.2 is on the Centos9 stream and the build with the Nutanix 20+ guest bug is on it.

PR #384, opened 2022-09-15
2099925: Drop support for RHEVM on RHEL 9

jirihnidek:
* BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2099925
* Card ID: ENT-5097
* Support for RHEVM was removed from code, unit tests and documentation
* Note: It is still possible to report host-guess mapping for RHEV, but libvirt has to be used.


wottop:
There is still a RHEVM reference in virt-who-config.5. Otherwise, ack.

wottop:
ack, will merge when first bug in main is in a build and confirmed.

PR #388, opened 2022-10-14
Added configuration for rhel-9.0 into releasers.conf

jirihnidek:
* Extended releasers.conf file to be able to trigger new build easily with 'tito release rhel-9.0' for this branch


wottop:
ack

PR #391, opened 2023-01-25
2158710: Migrated virt-who.conf, when necessary

jirihnidek:
* BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2158710
* Card ID: RHELPLAN-144095
* Do not try to update virt-who.conf file, when there is nothing new to add to this file. It means, there is not environment variable defined in sysconfig file.
* Added some doc string and type hints to describe the process
* Moved license block to comment
* Added doc string for the module describing purpose of the module


m-horky (approved):
LGTM

PR #396, opened 2023-04-13
spec: drop support for old systems/bits 

ptoscano:
Python 3.6 is the required Python version in `main`, hence drop the support for Python 2, and all the distributions that do not support Python 3; this means dropping the support for:
- RHEL older than 8
- Fedora older than 36 (older versions are EOL)

In addition to that:
- drop outdated variables and comments
- drop no more needed "six" dependency
- assume python3-libvirt naming everywhere


virt-who.spec
```diff
@@ -87,47 +74,25 @@ install -m 644 virt-who-zsh %{buildroot}/%{_datadir}/zsh/site-functions/_virt-wh
...
 
 %files
 %doc README.md LICENSE README.hyperv
 %{_bindir}/virt-who
 %{_bindir}/virt-who-password
 %{python_sitelib}/*
-%if %{use_systemd}
 %{_unitdir}/virt-who.service
-%else
-%{_sysconfdir}/rc.d/init.d/virt-who
```
jirihnidek, line 154: It would be also good to remove `virt-who-initscript` and installation step in `./setup.py` file.

jirihnidek (changes requested):
LGTM and thanks. I have one small request.

jirihnidek (approved):
LGTM :+1: Thanks for update

PR #397, opened 2023-04-19
Nutanix: fix issue with prism_central option

jirihnidek:
* When wrong value for prism_central was set, then virt-who
  was terminated with traceback
* Default value is set to `False`
* When wrong value is provided for boolean option, then print
  this value in message of Error and print default value
  in log message


m-horky (approved):
ACK

PR #398, opened 2023-04-20
Add support for GitHub Actions

jirihnidek:
* Added two jobs: pytest and stylish
* pytest is run on three containers: Fedora latests,
  Fedora rawhide and CentOS Stream 9
* flake 8 and rpmlint are used for checking style
* These two actions are triggered by pull requests
* Removed support for Jenkins


m-horky (approved):
LGTM

.github/workflows/pytest.yml
```diff
@@ -0,0 +1,47 @@
...
+            pytest_args: ''
+          - name: "Fedora Rawhide"
+            image: "fedora:rawhide"
+            pytest_args: ''
+
+    runs-on: ubuntu-latest
+    container:
+      image: ${{ matrix.image }}
+      volumes:
+        - /tmp:/tmp
```
ptoscano: this is not needed

.github/workflows/pytest.yml
```diff
@@ -0,0 +1,47 @@
...
+      - name: "Checkout repository"
+        uses: actions/checkout@v3
+
+      - name: "Run container-pre-test.sh"
+        run: |
+          bash scripts/container-pre-test.sh
+
+      - name: "Run pytest"
+        env:
+          SUBMAN_TEST_IN_CONTAINER: "1"
```
ptoscano: definitely not needed ;)

.github/workflows/stylish.yml
```diff
@@ -0,0 +1,34 @@
...
+    container:
+      image: fedora:latest
+
+    steps:
+    - name: Base setup
+      run: |
+        dnf --setopt install_weak_deps=False install -y \
+            git-core \
+            python3-flake8 \
+            python3-pip \
```
ptoscano: this is not needed, since black is not installed by an action

.github/workflows/pytest.yml
ptoscano, line 1: minor nit: missing newline at the end of the file

.github/workflows/stylish.yml
ptoscano, line 1: minor nit: missing newline at the end of the file

scripts/container-pre-test.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+dnf install -y libnl3-devel python3-libvirt python3-dateutil python3-setuptools python3-pip dnf-plugins-core \
+    python3-requests python3-cryptography python3-subscription-manager-rhsm subscription-manager
+dnf builddep -y virt-who.spec
```
ptoscano, line 5: i'd rework these in three steps:
- install `dnf-plugins-core`, so `dnf builddep` works
- `builddep` from the spec file, to install all the BuildRequires from the spec file automatically
- manual install of only the test-only packages

also, when installing stuff, please disable the weak dependencies (recommends), so we don't get unnecessary packages (makes the installation faster, and avoid side effects)

ptoscano (changes requested):
It is generally good, I left some notes of few things to improve/simplify (nothing problematic).

The problem I see are the requirements in `requirements.txt`: most of them are either outdated, useless, or conflicting with system stuff (we definitely don't want to install sub-man like this, as we haven't basically tested that for ages). Let me help here and cleanup those before we land this, so we can test more easily; I'll open a PR for this.

ptoscano:
Also, forgot to mention two things:
- in the commit that adds the actions, please mention ENT-5574
- a good idea IMHO can be adding the dependabot automatic update for the actions; I proposed it for sub-man (https://github.com/candlepin/subscription-manager/pull/3249), so in case we want it here too, I suggest to cherry-pick my commit there to this PR as well

scripts/container-pre-test.sh
jirihnidek: Problem is that `dnf builddep` does not work as expected for some reason. Many packages are just not installed and it is necessary to install them manually. I would like to depend only on `dnf builddep`, but it is not possible ATM.

ptoscano: which ones? anything I can check to help with this?

ptoscano:
> The problem I see are the requirements in `requirements.txt`: most of them are either outdated, useless, or conflicting with system stuff (we definitely don't want to install sub-man like this, as we haven't basically tested that for ages). Let me help here and cleanup those before we land this, so we can test more easily; I'll open a PR for this.

This is now #401.

scripts/container-pre-test.sh
jirihnidek: I mean I don't want to fix issue with `dnf builddep` as part of this PR.

ptoscano: well, if you don't even explain what the problem is, hardly people can help you...

PR #400, opened 2023-04-21
Removed support for Docker

jirihnidek:
* Removed support for Docker, because nobody uses this and there
  were definitions for old versions of CentOS not supported in main branch
* Cleaned `Makefile` a little


ptoscano (changes requested):
looks good, please also remove `.dockerignore` too

ptoscano (approved):

PR #401, opened 2023-04-27
Drop outdated/unused requirements

ptoscano:
- drop suds: now locally vendored
- drop subscription-manager: installing it from sources using pip is not something tested upstream anymore, so assume it is installed as system library
- drop iniparse & m2crypto: not used by virt-who (anymore?), likely needed by subscription-manager


jirihnidek (changes requested):
LGTM, but this PR needs rebase.

PR #403, opened 2023-04-28
spec: few more more cleanups

ptoscano:
- drop unused/unneeded `python3-setuptools` and `openssl` runtime dependencies
- drop no more needed `Group` tag
- drop old cleanup on `%install`
- hardcode/inline Python 3 variables


jirihnidek (approved):
LGTM :+1: Thanks

PR #404, opened 2023-05-16
Updated releaser configuration

jirihnidek:
* Added releaser configuration for rhel-8.9
* Updated Fedora section to support currently supported version
  of Fedora

### DIFF
--- a/rel-eng/releasers.conf
+++ b/rel-eng/releasers.conf
@@ -1,6 +1,6 @@
 [fedora]
 releaser = tito.release.FedoraGitReleaser
-branches = main f35 f34
+branches = main f37 f38
 
 [rhel-8.5]
 releaser = tito.release.DistGitReleaser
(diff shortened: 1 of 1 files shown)


rel-eng/releasers.conf
```diff
@@ -25,3 +25,11 @@ required_bz_flags = release+
 # Change this if you wish to use a placeholder "rebase" bug if none
 # are found in the changelog.
 placeholder_bz = 
+
+[rhel-8.9]
+releaser = tito.release.DistGitReleaser
+branches = rhel-8.9.0
```
ptoscano: 8.9+ follow the "stream" model already in use in el9, so you need to add a c8s releaser; see the equivalent for sub-man i added recently: https://github.com/candlepin/subscription-manager/commit/9a55e5348f5dcc2377857964f26896c9574f8037

ptoscano (changes requested):

rel-eng/releasers.conf
jirihnidek: Oh, I forgot about this. Good point.

wottop:
ack

ptoscano (approved):
thanks!

PR #405, opened 2024-01-18
fix: Use ConfigParser and not SafeConfigParser

jirihnidek:
* Card ID: CCT-263
* The `SafeConfigParser` was deprecated in Python 3.2. This class was renamed to `ConfigParser`
* Fixed some unit tests


m-horky (approved):
LGTM

PR #406, opened 2024-02-16
Migrate to SPDX license

jirihnidek:
* Changed string definition of license in .spec file. Licenses remain the same. It is still GPL for virt-who and LGPL for suds.


thrix:
Testing with fixes in https://gitlab.com/testing-farm/tests/-/merge_requests/36/diffs

PR #409, opened 2024-02-26
Change /var/run -> /run

opoplawski:
systemd complains:
```
systemd[1]: /usr/lib/systemd/system/virt-who.service:7: PIDFile= references a path below legacy directory /var/run/, updating /var/run/virt-who.pid → /run/virt-who.pid; please update the unit file accordingly.
```

### DIFF
--- a/virt-who.service
+++ b/virt-who.service
@@ -4,7 +4,7 @@ After=libvirtd.service network.target
 
 [Service]
 Type=notify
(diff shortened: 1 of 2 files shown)


jirihnidek (approved):
Thanks for the PR. :+1: You are right. We use legacy directory and it should be changed.

PR #411, opened 2024-05-15
feat: Changed release configuration for RHEL 10

jirihnidek:
* Removed all rhel-9 release configurations
* Updated Fedora configurations
* Added CentOS Stream 10 configuration
* Removed CentOS Stream 9 configuration
* Added configuration for RHEL 10 (main)

### DIFF
--- a/rel-eng/releasers.conf
+++ b/rel-eng/releasers.conf
@@ -1,32 +1,11 @@
 [fedora]
 releaser = tito.release.FedoraGitReleaser
-branches = main f36 f35
+branches = main f39 f40
 
-[centos-9-stream]
+[centos-10-stream]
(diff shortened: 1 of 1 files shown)


rel-eng/releasers.conf
```diff
@@ -1,32 +1,11 @@
...
-placeholder_bz =
-
-[rhel-9.2]
-releaser = tito.release.DistGitReleaser
-branches = rhel-9.2.0
-required_bz_flags = release+
-# Change this if you wish to use a placeholder "rebase" bug if none
-# are found in the changelog.
-placeholder_bz =
+branches = rhel-10-main
```
ptoscano, line 11: note that this will not actually used in practice, since updating the downstream `rhel-10-main` will always go through c10s

PR #412, opened 2024-07-01
fix: Added missing %{dist} to virt-who.spec

jirihnidek:
* Card ID: RHEL-44896
* The dist was removed by tito in previous commit


m-horky (approved):
LGTM

PR #413, opened 2024-07-01
fix: Replace all warn() with warning()

jirihnidek:
* The warn() has been deprecated for very long time and it just become not available
   in the newest version of Python 3.13
   https://docs.python.org/3.13/whatsnew/3.13.html#logging
* Fixed one unit test


m-horky (approved):
LGTM

PR #414, opened 2024-07-02
feat: Run Github CI action py.test every month

jirihnidek:
* Run unit tests every month to catch new regressions just in time


m-horky (approved):
LGTM, good idea

PR #415, opened 2024-07-31
fix: Try to convert ESX VM UUID after hw upgrade

jirihnidek:
* Card ID: CCT-606
* Added more debug prints
* Changed the logic a little bit. Do not convert the UUID
  from big-endian to little-endian, when
  acpi.smbiosVersion2.7: FALSE

### DIFF
--- a/virtwho/virt/esx/esx.py
+++ b/virtwho/virt/esx/esx.py
@@ -326,13 +326,37 @@ class Esx(virt.Virt):
         From: 78563412-AB90-EFCD-1234-567890ABCDEF
         To:   12345678-90AB-CDEF-1234-567890ABCDEF
         """
+        smbios_version_27 = None
         s = vm['config.uuid']
-        if 'config.version' not in vm:
-            return s
-        version = int(vm['config.version'].split('-')[1])
-        if (version >= 13):
-            return s[6:8] + s[4:6] + s[2:4] + s[0:2] + "-" + s[11:13] + s[9:11] + "-" + s[16:18] + s[14:16] + s[18:]
+
+        # Try to do some heuristics for one corner case described in RHEL-45636
+        if "config.extraConfig" in vm:
+            extra_config = vm["config.extraConfig"]
+            for item in extra_config.OptionValue:
+                if item.key == "acpi.smbiosVersion2.7":
+                    if item.value == "TRUE":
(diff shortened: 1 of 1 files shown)


virtwho/virt/esx/esx.py
```diff
@@ -326,11 +326,29 @@ def getVmUuid(self, vm):
...
         To:   12345678-90AB-CDEF-1234-567890ABCDEF
         """
+        smbios_version_27 = None
         s = vm['config.uuid']
-        if 'config.version' not in vm:
-            return s
-        version = int(vm['config.version'].split('-')[1])
-        if (version >= 13):
+
+        # Try to some crazy heuristics
```
cnsnyder: Non nacking comment: I think this sentence does not need the "to".
"Try to some crazy heuristics" -> "Try some crazy heuristics".

virtwho/virt/esx/esx.py
```diff
@@ -326,11 +326,29 @@ def getVmUuid(self, vm):
...
+                    if item.value == "TRUE":
+                        smbios_version_27 = True
+                    elif item.value == "FALSE":
+                        smbios_version_27 = False
+
+        self.logger.debug(f"ESX acpi.smbiosVersion2.7: {smbios_version_27}")
+
+        version = 0
+        if 'config.version' in vm:
+            version = int(vm['config.version'].split('-')[1])
```
cnsnyder, line 346: Are all config.version values of the same form? Can we be sure this won't result in an IndexError if we attempt to access [1] on the result of the split, if in case the format is different?

cnsnyder (commented):
Looks good to me overall, just a few questions to confirm before we proceed!

virtwho/virt/esx/esx.py
jirihnidek: Updated comment to be more descriptive.

virtwho/virt/esx/esx.py
jirihnidek: I do not think that this could happen. This code has worked for very long time and nobody has complained.

cnsnyder (commented):
Looks good to me, going to try a scratch build prior to merging! I'll update with anything we learn from it.

cnsnyder:
LGTM! Thank you!

PR #421, opened 2025-02-06
fix: Report correct guest UUID, when VM Generation 2 is used

jirihnidek:
* Card ID: CCT-676
* When VM is created as Generation 2 in the setup wizard of Hyper-V hypervisor, then hypervisor reports UUID of given VM using little-endian. Thus it is not necessary to convert this UUID as we do for Generation 1.
* We solve this issue by extending SQL query. We also ask for VirtualSystemSubType, which contains version of VM generation (`"Microsoft:Hyper-V:SubType:1"` or `"Microsoft:Hyper-V:SubType:2"`). When VM is generation 1, then we convert UUID from big-endian to little-endian as we have did it in the past. When generation is 2, then we do nothing with guest UUID.
* Changed code around little bit, because there was variable `uuid` used for different purpose and it was confusing.
* TODO: `hyperv.py` deserves more refactoring
* More information about generation 1 or 2 on Hyper-V could be found e.g. here:
  * https://learn.microsoft.com/en-us/windows-server/virtualization/hyper-v/get-started/create-a-virtual-machine-in-hyper-v
  * https://learn.microsoft.com/en-us/windows-server/virtualization/hyper-v/plan/should-i-create-a-generation-1-or-2-virtual-machine-in-hyper-v

### DIFF
--- a/virtwho/virt/hyperv/hyperv.py
+++ b/virtwho/virt/hyperv/hyperv.py
@@ -272,12 +272,15 @@ class HyperVSoap(object):
             "Content-Type": "application/soap+xml;charset=UTF-8"
         }
         try:
+            self.logger.debug("Trying to POST request to Hyper-V")
+            # self.logger.debug(body)
             response = self.connection.post(self.url, body, headers=headers)
         except requests.RequestException as e:
             raise HyperVException("Unable to connect to Hyper-V server: %s" % str(e))
 
         if response.status_code == requests.codes.ok:
             self.logger.debug(f'Received valid response from Hyper-V server: {response.status_code}')
+            # self.logger.debug(response.text)
             return response.content
         elif response.status_code == 401:
             raise HyperVAuthFailed("Authentication failed")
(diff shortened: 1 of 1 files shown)


virtwho/virt/hyperv/hyperv.py
```diff
@@ -427,17 +434,41 @@ def connect(self):
...
+        try:
+            vmms_major_ver = int(vmms_version.split(".")[0])
+        except (ValueError, KeyError):
+            self.logger.error(f"Wrong format of Hyper-V vmms version: {vmms_version}")
+            vmms_major_ver = 0
+
+        if gen_version is None or gen_version == self.GEN_VERSION_1:
+            return s[6:8] + s[4:6] + s[2:4] + s[0:2] + "-" + s[11:13] + s[9:11] + "-" + s[16:18] + s[14:16] + s[18:]
+        elif gen_version == self.GEN_VERSION_2:
+            if vmms_major_ver <= 10:
```
jirihnidek: We need to know the right value of this threshold. The value `10` is only estimation.

jirihnidek:
Closing, because this is not correct way how to deal with endianess of VM UUID. When candlepin is not able to find VM UUID using little-endian format, then it tries to find it using big-endian format. Thus, it does not make any sense to try to fix it on client side.

PR #417, opened 2025-01-15
fix: [1.31] Try to convert ESX VM UUID after hw upgrade

jirihnidek:
* Original PR: #415
* Backport to 1.31 branch
* Added more debug prints
* Changed the logic a little bit. Do not convert the UUID
  from big-endian to little-endian, when
  acpi.smbiosVersion2.7: FALSE

PR #424, opened 2025-05-15
fix: Revert [1.31] Try to convert ESX VM UUID after hw upgrade

jirihnidek:
* Card ID: CCT-606
* Card ID: CHAINSAW-427
* Card ID: TEAMNADO-7550
* Original PR was: https://github.com/candlepin/virt-who/pull/417
* This reverts commit 454887a07d42042e94e55e82699095f14a870586.
* Commit 454887a07d42042e94e55e82699095f14a870586 was not necessary, because when candlepin does not find VM using little-endian format, then it tries to find VM using big-endian format. Thus, change in commit 454887a07d42042e94e55e82699095f14a870586 was not necessary.
* It seems that customer issues have been fixed server side. See related cards.


m-horky (approved):
ACK

PR #427, opened 2026-03-20
fix(docks): fix duplicate numbering in virt-who man page

Peto-RH:
Correct "status mode" numbering from 3. to 4. in the USAGE section.


jirihnidek (approved):
LGTM

PR #428, opened 2026-04-13
fix: Handle hypervisor_id option properly

jirihnidek:
* Card ID: CCT-2241
* When `hypervisor_id` is set to hostname and Nutanix does not provide `name` property in host object, then fallback to `uuid` for `hypervisor_id`

### DIFF
--- a/virtwho/virt/ahv/ahv.py
+++ b/virtwho/virt/ahv/ahv.py
@@ -104,7 +104,6 @@ class Ahv(virt.Virt):
             Dictionary with host-guest mapping
         """
         mapping = {'hypervisors': []}
-        hypervisor_id = None
 
         host_uvm_map = self._interface.build_host_to_uvm_map()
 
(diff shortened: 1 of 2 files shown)


virtwho/virt/ahv/ahv.py
```diff
@@ -113,15 +112,18 @@ def get_host_guest_mapping_v3(self):
...
+                try:
                     hypervisor_id = host['name']
-
-            except KeyError:
-                self.logger.debug("Host '%s' doesn't have hypervisor_id property", host_uuid)
-                continue
+                    self.logger.debug("Using hypervisor_id: hostname: %s", hypervisor_id)
+                except KeyError:
+                    self.logger.warning("Host '%s' doesn't have 'name' property", host_uuid)
+                    self.logger.warning("Falling back to default hypervisor_id: uuid")
```
cnsnyder, line 126: Non-nacking comment: Since this is repeated is it worth having a helper function to do this?

cnsnyder (approved):
Overall LGTM, just one question which I don't think should block merge.

virtwho/virt/ahv/ahv.py
jirihnidek: Well, the fact that we handle two versions of API in the same code (`ahv.py`) has been reason for many confusions. We also do not know yet, if the behavior will be the same in both versions of API. Thus, I would keep it as it is now.

PR #430, opened 2026-07-20
feat(hyperv): add auth_method, kerberos_keytab, kerberos_principal config options

jlocash:
Add Kerberos authentication support to the HyperV config section.

New config keys on HypervConfigSection:
- `auth_method` (`basic`|`kerberos`, default: `basic`)
- `kerberos_keytab` (optional filesystem path to a keytab file)
- `kerberos_principal` (optional Kerberos principal string)

Behavior:
- auth_method=basic: preserves existing behavior; username and password remain required.
- auth_method=kerberos: username and password become optional. If provided, a warning is logged that they are ignored.
- kerberos_keytab/kerberos_principal set with auth_method=basic are removed with a warning.
- kerberos_keytab validates the file exists and is readable.
- Invalid auth_method values produce a clear validation error.

NOTE: These changes do not fully implement kerberos authentication in virt-who. Rather, this PR includes only changes to the config file. As such, there is not an easy way to test these changes, hence the feature branch. I will follow up with another PR with the remaining work

Assisted-by: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/tests/test_config_section_hyperv.py
+++ b/tests/test_config_section_hyperv.py
@@ -21,8 +21,25 @@ from __future__ import print_function
 Test validating of HypervConfigSection
 """
 
+import os
+import tempfile
+
+from functools import wraps
+
 from base import ConfigSectionValidationTests, TestBase
+from virtwho.config import ValidationState
 from virtwho.virt.hyperv.hyperv import HypervConfigSection
+
+
+def with_named_tempfile(func):
+    @wraps(func)
+    def inner(*args, **kwargs):
+        f = tempfile.NamedTemporaryFile(delete=False)
+        try:
+            return func(*args, f.name, **kwargs)
+        finally:
+            os.unlink(f.name)
+    return inner
 
 
 class TestHyperVConfigSection(ConfigSectionValidationTests, TestBase):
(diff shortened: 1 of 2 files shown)


tests/test_config_section_hyperv.py
```diff
@@ -52,3 +53,232 @@ class TestHyperVConfigSection(ConfigSectionValidationTests, TestBase):
...
+        self.assertEqual(config.state, ValidationState.VALID)
+
+    def test_auth_method_invalid_value(self):
+        """Invalid auth_method yields a clear validation error."""
+        values = dict(self.VALID_CONFIG, auth_method='ntlm')
+        config = self.CONFIG_CLASS.from_dict(values, "test", None)
+        messages = config.validate()
+        self.assertEqual(config.state, ValidationState.INVALID)
+        error_messages = [msg for msg in messages if msg[0] == 'error']
+        self.assertTrue(len(error_messages) > 0)
```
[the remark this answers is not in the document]
jlocash: Fair enough, suggestion applied

virtwho/virt/hyperv/hyperv.py
```diff
@@ -48,12 +48,95 @@ class HypervConfigSection(VirtConfigSection):
...
+        self.add_key('auth_method', validation_method=self._validate_auth_method, default='basic')
+        self.add_key('kerberos_keytab', validation_method=self._validate_kerberos_keytab)
+        self.add_key('kerberos_principal', validation_method=self._validate_kerberos_principal)
+
+    def _validate_auth_method(self, key):
+        result = None
+        try:
+            value = self._values[key]
+        except KeyError:
+            return result
```
cnsnyder: Could this instead be (for simplicity):
```python
if key not in self._values: return None
value = self._values[key]
```
I'd suggest the same where these four lines repeat to reduce verbosity.

tests/test_config_section_hyperv.py
```diff
@@ -52,3 +53,233 @@ class TestHyperVConfigSection(ConfigSectionValidationTests, TestBase):
...
+        del values['password']
+        config = self.CONFIG_CLASS.from_dict(values, "test", None)
+        messages = config.validate()
+        self.assertEqual(config.state, ValidationState.INVALID)
+        self.assertTrue(any('kerberos_keytab' in msg[1] and msg[0] == 'error' for msg in messages))
+
+    def test_kerberos_keytab_ignored_with_basic(self):
+        """kerberos_keytab with auth_method=basic produces a warning and is removed."""
+        import os
+        import tempfile
```
cnsnyder: I'd suggest just adding these to the imports at the top of the test file. Any benefit to importing these directly in each of the tests that use them?

Also we could add a utility decorator to make a few of these tests a lil simpler:
```python
def with_named_tempfile(func):
    def inner(*args, **kwargs):
        import os
        import tempfile
        f = tempfile.NamedTemporaryFile(delete=False)
        try:
            return func(*args, f.name, **kwargs)
        finally:
            os.unlink(f.name)
    return inner
```

Usage would be like:
```python
@with_named_tempfile
def test_kerberos_keytab_ignored_with_basic(self, keytab_path):
# Rest of test here can use keytab_path as it already does, no need to unlink file after
```

tests/test_config_section_hyperv.py
```diff
@@ -52,3 +53,233 @@ class TestHyperVConfigSection(ConfigSectionValidationTests, TestBase):
...
+        self.assertEqual(config['auth_method'], 'basic')
+
+    def test_auth_method_kerberos_valid_without_password(self):
+        """auth_method=kerberos is valid without password."""
+        values = dict(self.VALID_CONFIG, auth_method='kerberos')
+        del values['password']
+        config = self.CONFIG_CLASS.from_dict(values, "test", None)
+        config.validate()
+        self.assertEqual(config.state, ValidationState.VALID)
+        self.assertEqual(config['auth_method'], 'kerberos')
```
cnsnyder, line 105: Curious why this line is here but not in the tests 'test_auth_method_kerberos_valid_without_username'
nor
'test_auth_method_kerberos_valid_without_username_and_password'?

I guess I'd either rather it be in all of them if we're trying to show that the auth_method stays set to 'kerberos' when specifically set to it regardless of presence or absence of username / password.

cnsnyder (changes requested):
Overall looks good to me! Just a few things for discussion below! I don't think any are worth blocking this PR long on but if they are considered useful suggestions, I'd vote we take them and proceed!

tests/test_config_section_hyperv.py
jlocash: ~~Well, it doesn't really need to be here either. It's already covered in `test_auth_method_wired_through_config`. I suppose we could be redundant and add the checks elsewhere, but it shouldn't be necessary~~

Actually, redundancy is probably good for these tests, since I call out auth_method right in the docstring. I've added that assertion to the other tests where it was missing

cnsnyder (approved):
LGTM! Thank you!

PR #431, opened 2026-08-04
feat(hyperv): implement kerberos session auth 

jlocash:
This PR is a follow-up to #430 and wires up the new kerberos configuration options.

First, these changes add `python3-requests-gssapi` as a required dependency in the spec file. This package provides the `HTTPSPNEGOAuth` session auth used to authenticate via kerberos to HyperV. 

As for configuring kerberos auth, there are a few different ways:
- When `auth_method` is set to `kerberos`, virt-who will try and construct an `HTTPSPNEGOAuth` object. Otherwise, basic auth will be used
- If `kerberos_princpal` is unset, then the default principal will be used from the given keytab
- If `kerberos_keytab` (file path) is unset, the default credential cache will be used
- The two above values are used to construct a `gssapi.Credentials` object used by `HTTPSPNEGOAuth`.

Card ID: CCT-2189

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -12,3 +12,11 @@ files_to_sync:
 upstream_package_name: virt-who
 # downstream (Fedora) RPM package name
 downstream_package_name: virt-who
+
+jobs:
+  - job: copr_build
+    trigger: pull_request
+    targets:
+      - centos-stream-10
+      - fedora-all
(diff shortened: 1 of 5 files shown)


DuckBoss (approved):
LGTM, pytest failures are pre-existing and should be resolved in a separate PR.

PR #432, opened 2026-08-05
fix: handle str message decode in ESX Request/Reply unicode method

DuckBoss:
Request.__unicode__() and Reply.__unicode__() unconditionally called .decode("raw_unicode_escape") on self.message, which fails with AttributeError on Python 3.14+ when message is a str rather than bytes. This broke 24 suds transport tests on Fedora latest (Python 3.14) where pytest's log capture eagerly formats the log arguments.

Guard .decode() with an isinstance(message, bytes) check so both str and bytes messages are handled correctly.

Assisted-by: Cursor

### DIFF
--- a/tests/test_suds_transport.py
+++ b/tests/test_suds_transport.py
@@ -1,0 +1,42 @@
+from __future__ import print_function
+
+import unittest
+
+from virtwho.virt.esx.suds.transport import Request, Reply
+
+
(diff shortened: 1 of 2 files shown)


DuckBoss:
Fixes pytest failures that show up in this PR: https://github.com/candlepin/virt-who/pull/431

jlocash (approved):
LGTM, thanks for the PR!

PR #433, opened 2026-08-19
feat(kubevirt): support namespace-scoped API access

jlocash:
Add a `namespace` config option so kubevirt virtual machine queries can be restricted to a single namespace instead of requiring cluster-wide access. This supports multi-tenant clusters where virt-who is only granted namespace-scoped RBAC permissions.

Node listing (`/api/v1/nodes`) always requires cluster-wide access and is unaffected by `namespace`. Since that access isn't guaranteed in namespace-scoped deployments, a `hypervisor_id_fallback` option is also added:
- By default, if node listing returns 403 Forbidden, virt-who treats it as a fatal error.
- Setting `hypervisor_id_fallback=true` instead falls back to identifying hosts by the node name reported on each virtual machine (VMI). In this degraded mode, hypervisor facts normally derived from node data (CPU count, hypervisor version, system uuid) are not reported.

`statusConfirmConnection` follows the same behavior, so virt-who --status accurately reflects whether a real run would succeed or fail.

Card ID: CCT-2379

Assisted-by: Claude Sonnet 5 <noreply@anthropic.com>


template.conf
```diff
@@ -17,6 +17,8 @@
 #hypervisor_id=      ; how will be the hypervisor identified, one of: uuid, hostname, hwuuid
 #kubeconfig=         : path to kubernetes configuration file - kubevirt specific option
 #kubeversion=        : used to override kubevirt api version - kubevirt specific option
+#namespace=          : restrict kubevirt queries to a single namespace - kubevirt specific option
+#hypervisor_id_fallback= ; set to true to identify hosts by node name if node listing is forbidden (403) - kubevirt specific option
```
jlocash, line 21: I was debating whether or not to remove this option and just make it the default behavior when `namespace` is set. Logically, there is nothing preventing one from setting the namespace while also using a cluster-scoped SA, so in the case there's something else wrong with the config, I didn't want virt-who silently ignoring the configured value for `hypervisor_id`.

tests/test_kubevirt.py
```diff
@@ -251,6 +252,122 @@ def test_getHostGuestMapping_with_hm(self):
             result = kubevirt.getHostGuestMapping()['hypervisors'][0]
             self.assertEqual(expected_result.toDict(), result.toDict())
 
+    def test_getHostGuestMapping_nodes_forbidden(self):
```
jirihnidek: Can you please add doc string to all tests? I am trying to understand what is this methods testing and I do not know.

cnsnyder (approved):
LGTM one minor thing though: why two commits? Seems worth a squash to me!

template.conf
cnsnyder: Seems ok to me to keep, though I'm fairly ambivalent on it.

PR #436, opened 2026-09-21
ci(packit): fix RPM version parsing from git tags

jlocash:
Our git tags are named virt-who-<version>-<release>, but Packit's default upstream_tag_template treats the whole tag as the version. This produced mangled RPM names like
virt-who-virt.who.1.32.1.1-1....el10.noarch on PR copr builds.

Set upstream_tag_template so Packit extracts just the version portion.


jirihnidek (approved):
LGTM

